PR #8, opened 2023-09-15
Delete functionality

thomasyeung687:
I added the ability to delete posts using NIP 9 and kind 5 events.
I noticed some issues with create product that I resolved. 
I also cleaned up the post-event.ts as there was alot of repeated code. Further optimizations can be made.


src/pages/api/nostr/post-event.ts
```diff
@@ -41,10 +41,10 @@ const parseRequestBody = (body: string) => {
     console.log("Missing or invalid property: publicKey");
     throw new Error('Invalid request data: missing or invalid property');
   }
-  if (!parsedBody.privkey || typeof parsedBody.privkey !== 'string') { 
-    console.log("Missing or invalid property: privateKey");
-    throw new Error('Invalid request data: missing or invalid property');
-  }
+  // if (!parsedBody.privkey || typeof parsedBody.privkey !== 'string') {
```
thomasyeung687, line 44: Some requests won't have a private key if they are signed by the extension. AKA a delete event that I just implemented

src/pages/api/nostr/post-event.ts
```diff
@@ -195,120 +195,65 @@ const PostEvent = async (req: NextApiRequest, res: NextApiResponse) => {
...
-        console.log('got event:', event);
-      });
-
-      await pool.publish(relays, signedEvent);
       
-      let events = await pool.list(relays, [{ kinds: [0, kind] }]);
-      let postedEvent = await pool.get(relays, {
-        ids: [event.id],
-      });
+      signedEvent = finishEvent(nip04Event, privkey);
```
thomasyeung687: I see finishEvent in all of these event types. Can we extract that out to somewhere? Also do we need all of these types? @calvadev

src/pages/components/product-form.tsx
```diff
@@ -311,76 +219,11 @@ const ProductForm = ({
                       type="text"
                       id="t"
                       name="t"
-                      value={formValues.t}
+                      value={getFormValue('t')}
                       placeholder="Optional"
                       onChange={handleChange}
                       className="w-full p-2 border border-gray-300 rounded"
                     />
-
```
thomasyeung687, line 319: Getting rid of this. If in the future we want to reimplement this stuff. we can reference previous commits

src/pages/api/nostr/post-event.ts
calvadev: This will only be accessed if the user signs in with a private key so we should keep it.

src/pages/api/nostr/post-event.ts
calvadev: We need finishEvent to be called independently since each event corresponding to a different kind is constructed differently. As for the other deleted code, really only `await pool.publish(relays, signedEvent);` needs to be added back. The sub calls just monitor the relay list to see when the event is posted to the relays and basically the same goes for pool.list and pool.get.

src/pages/api/nostr/post-event.ts
thomasyeung687: This throws an error when a user that signs in with an extension trys to post an event because the event won't contain his private key. Thus I removed it as the original logic is expecting everyone to have a private key in the request.

src/pages/api/nostr/post-event.ts
thomasyeung687: I fixed the ordering of the postevents function and pulled the finish events out. LMK if the logic makes sense for you. I think this makes the function alot cleaner and still the same functionality. Also the `await pool.publish(relays, signedEvent);` is after the if statements

calvadev (approved):

PR #9, opened 2023-09-21
qof improvements

ericspaghetti:
General qof fixes
* flatten repo structure, run npm commands from root folder
* get files to build ie no type errors & linting issues etc (left dm page alone as it's being worked on by @calvadev 
* add default relay, can view marketplace without having to go through login setup
* add delete product feature in product id page
* Use next/Image where appropriate instead of img (helps with optimization according to next)
* Use next/Link where appropriate instead of router.push (helps with optimization according to next)


ericspaghetti:
adding these in seperate PRs

PR #28, opened 2023-10-20
Updated sign in flow

calvadev:
- Added sign in option in navbar
- Added alerts to sign in if signIn doesn't exist in local storage
- Marketplace view is now the root url


thomasyeung687 (approved):
LGTM

PR #51, opened 2024-02-17
Share products

calvadev:
- Added ability to share links directly to product listings
- Removed automatic invoice generation
- Added meta tags for direct-to-product links
- Renamed usages of "checkout" to more accurately describe functionality


pages/components/display-product-modal.tsx
```diff
@@ -70,6 +68,25 @@ export default function DisplayProductModal({
...
+    // Check if the Web Share API is available
+    if (navigator.share) {
+      // Use the share API
+      await navigator.share(shareData);
+    } else {
+      // Fallback for browsers that do not support the Web Share API
+      navigator.clipboard.writeText(
+        `https://shopstr.store/listing/${productData.id}`,
+      );
+      alert("Listing URL copied to clipboard!");
```
thomasyeung687, line 86: I think we should move away from alerts, it makes the site seem sorta unpolished. Perhaps an animation of copied to clipboard would be sufficient

pages/components/invoice-card.tsx
```diff
@@ -11,23 +12,34 @@ import {
...
+export default function InvoiceCard({
   productData,
 }: {
   productData: ProductData;
 }) {
   const router = useRouter();
   const { pubkey, currency, totalCost } = productData;
   const pubkeyOfProductBeingSold = pubkey;
-  const { decryptedNpub, relays, mints } = getLocalStorageData();
+  const { npub, decryptedNpub, relays, mints } = getLocalStorageData();
```
thomasyeung687, line 40: there is a type error occuring for npub. Did the linter show this on your side?

pages/components/invoice-card.tsx
```diff
@@ -54,22 +66,21 @@ export default function CheckoutCard({
...
     const profileMap = profileContext.profileData;
     const profile = profileMap.has(decryptedNpub)
       ? profileMap.get(decryptedNpub)
       : undefined;
-    setName(
-      profile && profile.content.name
-        ? profile.content.name
-        : nip19.npubEncode(decryptedNpub),
-    );
+    if (typeof window !== "undefined") {
```
thomasyeung687: Still think this check is not needed. Am free to hop on a call to discuss

pages/components/invoice-card.tsx
```diff
@@ -11,23 +12,34 @@ import {
...
 }: {
   productData: ProductData;
 }) {
   const router = useRouter();
   const { pubkey, currency, totalCost } = productData;
   const pubkeyOfProductBeingSold = pubkey;
-  const { decryptedNpub, relays, mints } = getLocalStorageData();
+  const { npub, decryptedNpub, relays, mints } = getLocalStorageData();
+
+  const [paymentCard, setPaymentCard] = useState(false);
```
thomasyeung687: I think we need a better name for this cause the current name is ambiguous. Maybe try something like showPaymentQRCode or soemthing along those lines

pages/components/invoice-card.tsx
```diff
@@ -189,69 +200,121 @@ export default function CheckoutCard({
...
+            }
+          >
+            Message
+          </Button>
+          <Button
+            type="submit"
+            className={SHOPSTRBUTTONCLASSNAMES + " mt-3"}
+            onClick={() => {
+              let { signIn } = getLocalStorageData();
+              if (!signIn) {
```
thomasyeung687, line 229: !getLocalStorageData().signIn could be a way to make this cleaner. Or even turn signedIn to a useState

pages/components/invoice-card.tsx
calvadev: If you remove this it'll fail to load.

pages/components/display-product-modal.tsx
calvadev: We have to change all the copied to clipboard uses then and make an animation of some sort. Best to be done in a separate pr I think.

pages/listing/[[...productId]].tsx
```diff
@@ -4,21 +4,20 @@ import { SimplePool } from "nostr-tools";
...
+const Listing = () => {
   const router = useRouter();
   const [relays, setRelays] = useState([]);
   const [productData, setProductData] = useState<ProductData | undefined>(
     undefined,
   );
 
-  const { productId } = router.query;
-  const productIdString = productId ? productId[0] : "";
+  const productIdString = router.asPath.split("/").pop() || ""; // Extract productId from the actual URL
```
Comment (thomasyeung687) [inline pages/listing/[[...productId]].tsx]: Not sure if this is the correct way to use router as it was intended since there is the .query functionality

pages/components/invoice-card.tsx
calvadev: Updated it to invoiceCard

PR #59, opened 2024-02-25
Cashu redeem

calvadev:
- Added one-click Cashu token redemption
- Defaults to '@npub.cash' lnurl if user doesn't have Kind 0 lud16 set
- Added success and failure modals to indicate payment status

### DIFF
--- a/components/utility-components/redeem-button.tsx
+++ b/components/utility-components/redeem-button.tsx
@@ -31,14 +31,16 @@ function decodeBase64ToJson(base64: string): any {
 export default function RedeemButton({ token }: { token: string }) {
   const [lnurl, setLnurl] = useState("");
   const profileContext = useContext(ProfileMapContext);
-  const { npub, decryptedNpub, mints, relays } = getLocalStorageData();
+  const { npub, decryptedNpub, relays } = getLocalStorageData();
 
   const [openRedemptionModal, setOpenRedemptionModal] = useState(false);
   const [isPaid, setIsPaid] = useState(false);
   const [isCashu, setIsCashu] = useState(false);
   const [isSpent, setIsSpent] = useState(false);
   const [isRedeeming, setIsRedeeming] = useState(false);
+  const [wallet, setWallet] = useState<CashuWallet>();
   const [proofs, setProofs] = useState([]);
+  const [tokenMint, setTokenMint] = useState("");
   const [tokenAmount, setTokenAmount] = useState();
   const [formattedTokenAmount, setFormattedTokenAmount] = useState();
   const [redemptionChangeAmount, setRedemptionChangeAmount] = useState();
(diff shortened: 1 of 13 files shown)


components/utility-components/redeem-button.tsx
```diff
@@ -0,0 +1,169 @@
...
+      .catch((error) => {
+        console.error(error);
+      });
+  }, []);
+
+  useEffect(() => {
+    const decodedToken = decodeBase64ToJson(token);
+    const proofs = decodedToken.token[0].proofs;
+    setProofs(proofs);
+    const totalAmount = proofs.reduce(
```
thomasyeung687: change this to 
```
const totalAmount =
      Array.isArray(proofs) && proofs.length > 0
        ? proofs.reduce((acc, current) => acc + current.amount, 0)
        : 0;
```
so the error does not occur when entering a chat with in properly formed cashu token

components/utility/nostr-helper-functions.ts
```diff
@@ -347,7 +347,9 @@ export const getLocalStorageData = (): LocalStorageInterface => {
     if (
       mints === null ||
       mints[0] ===
-        "https://legend.lnbits.com/cashu/api/v1/4gr9Xcmz3XEkUNwiBiQGoC"
+        "https://legend.lnbits.com/cashu/api/v1/4gr9Xcmz3XEkUNwiBiQGoC" ||
+      mints[0] !==
+        "https://legend.lnbits.com/cashu/api/v1/AptDNABNBXv8gpuywhx6NV"
```
thomasyeung687: Im confused on this logic here. 
If item in mints is this or it isn't this. 

why not just 
mints === null || mints[0] !== "https://legend.lnbits.com/cashu/api/v1/AptDNABNBXv8gpuywhx6NV"

pages/settings/index.tsx
```diff
@@ -348,14 +348,14 @@ const SettingsPage = () => {
           </div>
         )}
 
-        <div className="flex h-fit flex-row justify-between bg-light-bg px-3 py-[15px] dark:bg-dark-bg">
+        {/* <div className="flex h-fit flex-row justify-between bg-light-bg px-3 py-[15px] dark:bg-dark-bg">
```
thomasyeung687: Is this supposed to be commented out?

components/utility-components/redeem-button.tsx
thomasyeung687: Also perhaps you have to look into why this cashu token for 0 sats was sent.

components/utility-components/redeem-button.tsx
```diff
@@ -0,0 +1,169 @@
...
+      ? sellerProfileMap.get(decryptedNpub)
+      : undefined;
+    setLnurl(
+      sellerProfile && sellerProfile.content.lud16
+        ? sellerProfile.content.lud16
+        : npub + "@npub.cash",
+    );
+  }, [profileContext]);
+
+  const formatter = new Intl.NumberFormat("en-GB", {
```
ericspaghetti: don't see this used

pages/settings/index.tsx
calvadev: Yes, atm restricting mint changes as it could cause tokens to be sent with no information on the mint. I'll add that functionality back in a separate pr.

ericspaghetti (approved):

PR #110, opened 2025-04-01
ui enhancement in landing page and product-card

YadlaMani:
This PR improves the UI of the landing page and product card by refining styles, optimizing layouts, and enhancing responsiveness. 
[image](https://github.com/user-attachments/assets/8363cb2c-bb1b-48d1-8419-08e84eb3d913)
image
image
image


calvadev:
Some nice touches, but I have a few comments:

- On the "Start Shopping" button, the icon isn't necessary. I believe the only place icons are located is in the wallet component, but it might be nice to have more consistent icons across all of the buttons, not just the "start shopping" one if it is added.

- The border outside of the product card components doesn't look quite right due to the way the images are configured within the card. For it to look more consistent the images would need to be resized to always fit the same size. I like the "featured products" addition, but it's more appropriate to call it "latest products."

- The why choose section looks really nice! I would just center the card titles ("permissionless commerce," "bitcoin native," and "privacy first" so it looks a bit cleaner.

- The "how it works" section also looks great; the only thing that should be fixed is that the images under each step should be on the same line, so either the text needs to be shrunk (not preferred) or the other three images need to have some margin added above them.

YadlaMani:
@calvadev will do this changes and also should I make different components for each section to make more readable and modular code ?

YadlaMani:
image
image
image
image

calvadev:
This looks great! One final touch-up: on the latest products section, if the images could be centered when resized that would be ideal. At least from this view the primary content of images of the product won't be out of frame (like in the sheep stickers listing).

calvadev:
oh also, the "active"/"sold" tag doesn't need to be displayed in the latest listings view.

YadlaMani:
@calvadev okay will be working on it and should I make components of the content here or it's fine like this

calvadev:
There are build issues coming up with this PR, but otherwise it looks great! Make sure to merge/rebase the changes from my latest PR. It looks big, but most of the changes are simply linting fixes.

PR #118, opened 2025-04-09
Enhance Navbar UI and Add Active State Highlighting

Abhi25jk:
I  improved the navigation bar UI by adding interactivity and better layout alignment. Some of the Key changes include:-

1) Added active state styling to navbar tabs (e.g., Marketplace, Orders, Wallet, etc.) for better navigation feedback.

2) Enhanced UI spacing and alignment to create a cleaner, more consistent look.

3) Improved cart icon display with dynamic update to show item count.

### Before
Navbar_Before

### After 
Navbar_After

### DIFF
--- a/components/nav-top.tsx
+++ b/components/nav-top.tsx
@@ -1,117 +1,84 @@
 import React, { useContext, useEffect, useState } from "react";
-import useNavigation from "@/components/hooks/use-navigation";
+import { useRouter } from "next/router";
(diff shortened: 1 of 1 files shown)


Abhi25jk:
Updated the Navbar , tried for the active navbar but it wasn't doing good to the UI

calvadev:
Will give this a look today!

calvadev:
If this added a similar indicator for the number of unread messages that would be great! Should be relatively simple to add.

PR #125, opened 2025-04-15
Fixed Web of Trust Minimum Follower Count Setting Not Persisting

GautamBytes:
**Summary**:
- We identified that the “Web of Trust” slider’s value was not updating as expected because it was implemented as an uncontrolled component using the defaultValue prop. This caused the slider to retain the initial default of 3 even after a change, and upon refresh the initialization function (getLocalStorageData) reverted the value to the default when a stored value was not properly detected.

- To address this for the current session, we refactored the slider as a controlled component (using the value prop) and removed the “Refresh to Apply” button so that changes update immediately and persist within the session. 

**Note**: However, on a full page refresh, the slider still resets to the default value, indicating that further investigation into the persistence mechanism in getLocalStorageData is needed for cross-session retention.

### DIFF
--- a/components/utility-components/shopstr-slider.tsx
+++ b/components/utility-components/shopstr-slider.tsx
@@ -1,63 +1,45 @@
 import { useState, useEffect, useContext } from "react";
-import { Button } from "@nextui-org/react";
 import { Slider } from "@nextui-org/react";
 import { useTheme } from "next-themes";
 import { FollowsContext } from "../../utils/context/context";
 import { getLocalStorageData } from "@/utils/nostr/nostr-helper-functions";
-import { SHOPSTRBUTTONCLASSNAMES } from "@/utils/STATIC-VARIABLES";
 
 const ShopstrSlider = () => {
   const { theme } = useTheme();
-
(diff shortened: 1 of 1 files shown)


GautamBytes:
@calvadev , Can you please review this approach and tell me if this fix is okay because i tried couple of ways to solve this with tweaks to already implemented logic but value changes to default value of 3 after each refresh (and we have kept refresh to apply button for the updation of new wot value( and thus it is not updated). seems like it starts a new session after reloading just like we see in nostr-profile section where we need to add a passphrase to keep current session going on (atleast when we are browsing).

GautamBytes:
@calvadev any Pov on this one ?

calvadev:
@GautamBytes If the WoT score still doesn't persist, then it's most likely an issue with how either the state variable is being handled upon load or how the value is being fetched from local storage. On my end it seems to work consistently, so it could also be an issue with not waiting on the follower lists to be fetched before modifying.

calvadev:
Resolved here: https://github.com/shopstr-eng/shopstr/commit/3e9bf559e5f401d8ad801d67ff5ecc6cad3bf5e7

PR #127, opened 2025-04-15
Prevent infinite recursion in authentication flow

Nj221102:
## Problem
The application encounters a "Maximum call stack size exceeded" error during sign-in. This happens because:

1. `loadSigner()` calls `setLocalStorageDataOnSignIn()`
2. `setLocalStorageDataOnSignIn()` updates localStorage and dispatches a storage event
3. The event listener calls `loadSigner()` again, creating an infinite loop

## Solution Explained

1. **Conditional localStorage Updates**: We only update localStorage when necessary (when data is missing or has changed).

2. **Custom Event with Flag**: We dispatch events with a special flag that lets our handler know which events it should ignore.

3. **Event Handler with Flag Check**: Our new handler function ignores self-triggered events but processes others normally.

4. **Initial Load**: We maintain the initial `loadSigner()` call to load authentication state when the component first mounts.

This solution breaks the infinite recursion while maintaining all the necessary functionality of the authentication system.

## Screenshot of issue
Screenshot 2025-04-16 at 12 48 25 AM

## Small changes
- updated caniuse-lite, to prevent this warning during build process
```
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
```

### DIFF
--- a/components/utility-components/nostr-context-provider.tsx
+++ b/components/utility-components/nostr-context-provider.tsx
@@ -11,10 +11,7 @@ import {
   NostrSigner,
 } from "@/utils/nostr/signers/nostr-signer";
 import { NostrManager } from "@/utils/nostr/nostr-manager";
-import {
-  getLocalStorageData,
-  setLocalStorageDataOnSignIn,
-} from "@/utils/nostr/nostr-helper-functions";
+import { getLocalStorageData } from "@/utils/nostr/nostr-helper-functions";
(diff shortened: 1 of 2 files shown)


Nj221102:
@calvadev Plz take a look.

calvadev:
LGTM

calvadev (approved):

PR #132, opened 2025-04-18
Added lint.yml to github workflow

GautamBytes:
**Summary**:
This adds a GitHub Actions workflow (.github/workflows/lint.yml) that automatically runs ESLint on all pages and components files and verifies Prettier formatting for every PR against main. If any lint or formatting issues are detected, the workflow posts a summary comment on the PR with guidance on how to fix them, ensuring consistent code quality across contributions.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,33 +1,41 @@
 name: ESLint & Prettier Check
+
 on:
-  pull_request:
-    branches: [ main ]
+  pull_request_target:
+    branches: [ main ]  
+
 permissions:
-  contents: read
-  pull-requests: write
-  issues: write
+  contents: read         
+  pull-requests: write   
+  issues: write         
+
 jobs:
   lint:
-    name: ESLint & Prettier
     runs-on: ubuntu-latest
+
     steps:
       - name: Check out code
         uses: actions/checkout@v4
         with:
+          token: ${{ secrets.GITHUB_TOKEN }}
           fetch-depth: 0
+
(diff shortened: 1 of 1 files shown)


GautamBytes:
@calvadev I've been trying to add the ESLint & Prettier workflow , but it's failing with a 'Resource not accessible by integration' error when trying to comment on PRs. This is a permissions issue. Could you please check the repository's workflow permissions under Settings > Actions > General and make sure 'Read and write permissions' is selected for workflow permissions? This would allow the workflow to post linting results as comments on pull requests.

calvadev:
I checked the permissions and I've seen that I'm actually not allowed to the read and write permissions for workflows at the moment (the setting is greyed out). I'm looking into how to add that.

calvadev:
@GautamBytes from what I've been able to find, you have to set an explicit permission set in order for it to work properly as such:

```
Add the permissions key to the job or the root of workflow (in this case it is applied to all jobs in the workflow that do not have their own permissions key) and assign the least privileges required to complete the task:

name: "My workflow"
permissions:
  contents: read
  pull-requests: write
or

jobs:
  my-job:
    permissions:
      contents: read
      pull-requests: write
```

GautamBytes:
Okay @calvadev , I will add it in the workflow file itself!

GautamBytes:
@calvadev , updated the workflow as per your suggestions . You can have a look!

calvadev:
LGTM!

calvadev (approved):

PR #136, opened 2025-04-24
Dark mode feature added and marketplace page UI improved

drstrox:
@calvadev I have added the dark mode feature onto the home page along with the improvement of the marketplace page.
Screenshot (39)

Screenshot (40)

### DIFF
--- a/components/home/marketplace.tsx
+++ b/components/home/marketplace.tsx
@@ -1,143 +1,74 @@
 import { MagnifyingGlassIcon } from "@heroicons/react/24/outline";
 import {
-  Button,
-  Chip,
   Select,
   SelectItem,
   SelectSection,
   Input,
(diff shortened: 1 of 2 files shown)


GautamBytes:
@drstrox While i haven't neither skimmed through your code and nor i am a mentor , but if you explore website the dark mode feature is already supported (settings > preferences > bottom  of page > options for dark/light/system). So , i think you can revert changes in pages/index.tsx . Regarding changes in marketplace.tsx , Calvadev would be reviewing it .

calvadev:
As Gautam said, dark mode is already supported under settings. I also see the updated view as not being very clean as compared to the previous UI, particularly with the color contrast, uneven drop-down menus, and squished-looking card view.

PR #143, opened 2025-04-27
order section improved

drstrox:
image

@calvadev


calvadev:
These changes are not consistent with the styling and design of the rest of the site. When making design changes you have to consider contrast with the entire application, not just one page.

drstrox:
Thanks for the feedback! I will take care of it while making changes in the design.

PR #150, opened 2025-05-22
Fix/pledge tag mapping

ritty65:


calvadev:
Also not relevant; closing.

PR #152, opened 2025-05-28
Added P2PK Time-Locked Cashu Tokens

GautamBytes:
In this PR , I have implemented the p2pk time locked cashu tokens as per NUT-11.

- Generate and parse P2PK secrets with absolute UNIX locktime and refund tags in Shop Settings
- Update TS types to reflect { locktime: number; refund?: string[] }
- Round-trip the locktime → locktimeDays in the shop form
- Compute and display “Locked for X days” in product and checkout cards from the new timestamp
- Enhance claim-button to handle both on-time spends and post-expiry refunds
- Align invoice components (cart-invoice-card, product-invoice-card) with the P2PK flow

### DIFF
--- a/components/cart-invoice-card.tsx
+++ b/components/cart-invoice-card.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import React, { useContext, useState, useEffect, useMemo } from "react";
 import { useRouter } from "next/router";
 import {
@@ -72,7 +71,6 @@ import {
   ShippingFormData,
   ContactFormData,
   CombinedFormData,
-  GetInfoResponse,
 } from "@/utils/types/types";
 
 export default function CartInvoiceCard({
@@ -1159,7 +1157,7 @@ export default function CartInvoiceCard({
 
       if (sellerAmount > 0) {
         const mint = new CashuMint(mints[0]!);
-        const info = (await mint.getInfo()) as unknown as GetInfoResponse;
+        const info = (await mint.getInfo()) as any;
         // NUT-11 support is advertised in `methods["11"].supported`
         if (!info.methods?.["11"]?.supported) {
           throw new Error("Mint does not support P2PK (NUT-11)");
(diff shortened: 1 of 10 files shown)


components/cart-invoice-card.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
calvadev: You can remove this and just delete 'GetInfoResponse' from the type import statement.

components/cart-invoice-card.tsx
```diff
@@ -71,6 +72,7 @@ import {
   ShippingFormData,
   ContactFormData,
   CombinedFormData,
+  GetInfoResponse,
```
calvadev: Delete this import to resolve the eslint error.

components/utility-components/checkout-card.tsx
```diff
@@ -122,20 +123,11 @@ export default function CheckoutCard({
 
   useEffect(() => {
     setIsFetchingShop(true);
-    if (
```
calvadev: This whole useEffect is removed in the main branch. Just rebase to resolve.

utils/types/types.ts
```diff
@@ -27,6 +27,13 @@ export interface ShopProfile {
       theme: string;
       darkMode: boolean;
     };
+    p2pk?: {
```
calvadev: This is a good approach, but it might make more sense to attach these values in the user Nostr profile rather than the shop profile as the shop profile is primarily for styling and not settings preferences. Another option is to create a settings event in the settings/preferences.tsx page that stores these p2pk values along with the other settings listed there. What do you think?

GautamBytes: I thought it would be good if merchant can enable timelock while putting details for shop itself. But yeah it makes more sense to have it on user nostr profile since payment preference option is also there. will update it accordingly

Also, since this will be new feature and it might be possible that some merchant can't understand p2pk timelock , so i can add  its info under FAQ's page also.

calvadev:
Superceded here: https://github.com/shopstr-eng/shopstr/commit/f9aa0606272348dd087471147d479703274040f6

PR #222, opened 2026-01-28
Fix banner image upload toString crash bug

raj-125-beep:
**### Fix shop banner / image upload crash and improve robustness**

What this PR does?

This PR fixes a runtime crash that occurred when uploading a shop banner or image, where the app threw:

Cannot read properties of undefined (reading 'toString')

The issue was caused by image upload flows not safely handling missing or undefined values during the upload → URL extraction process.

Key changes

Fixed image upload handling to prevent .toString() being called on undefined
Ensured image URLs are only processed once they are valid and present
Improved robustness of the shop profile image/banner upload flow
Prevented silent crashes during banner or profile image upload

Why this matters

Uploading a shop banner or profile image was completely broken for some users
The error surfaced at runtime and blocked shop profile updates
This fix makes image uploads reliable and failure-safe

How it was tested

Manually tested banner upload in Shop Profile
Verified profile image upload works without runtime errors
Confirmed no regressions in existing image display logic

### DIFF
--- a/utils/nostr/nostr-helper-functions.ts
+++ b/utils/nostr/nostr-helper-functions.ts
@@ -24,6 +24,8 @@ import {
   deleteEventsFromDatabase,
 } from "@/utils/db/db-client";
 import { newPromiseWithTimeout } from "@/utils/timeout";
+import { BLOSSOM_SERVER } from "@/constants/blossom";
+// Adding Blossom_server for extra checks!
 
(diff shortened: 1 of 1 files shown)


calvadev:
This doesn't use the Blossom spec conventions for file uploading, more servers will reject uploads than not.

Reference: https://github.com/hzrd149/blossom

Issue #224, opened 2026-02-14
[BUG] error: relation "failed_relay_publishes" does not exist


RiH-137:
### Reported Error

`GET /api/db/get-failed-publishes` was returning `500` with:

```text
error: relation "failed_relay_publishes" does not exist
code: 42P01
```

### Current Behavior

1. `get-failed-publishes` queried `failed_relay_publishes` before ensuring the table exists.
2. The same endpoint also joined a non-existent `events` table.
3. Failed publish tracking did not consistently store event payload (`event_data`) needed for retries.

### Expected Behavior

Database tables initialized successfully

### Steps to Reproduce

1. Added `failed_relay_publishes` table and indexes in DB schema/bootstrap.
2. Added migration support for `event_data` (`ALTER TABLE ... ADD COLUMN IF NOT EXISTS`).
3. Updated API endpoints to self-heal schema on request path where needed.
4. Removed invalid join to `events` table and fetched data directly from `failed_relay_publishes`.
5. Updated client and publish call-sites to send/store full failed event payload.

### Next.js Version

14.2

### Operating System

Windows 11

### Node.js Version

20

### Browser

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #225, opened 2026-02-14
fix/error: relation "failed_relay_publishes" does not exist

RiH-137:
…creation + migration support in DB

### Description
error: relation "failed_relay_publishes" does not exist
code: 42P01

### Resolved or fixed issue
#224

### Root Cause

1. `get-failed-publishes` queried `failed_relay_publishes` before ensuring the table exists.
2. The same endpoint also joined a non-existent `events` table.
3. Failed publish tracking did not consistently store event payload (`event_data`) needed for retries.

---

## Solution

1. Added `failed_relay_publishes` table and indexes in DB schema/bootstrap.
2. Added migration support for `event_data` (`ALTER TABLE ... ADD COLUMN IF NOT EXISTS`).
3. Updated API endpoints to self-heal schema on request path where needed.
4. Removed invalid join to `events` table and fetched data directly from `failed_relay_publishes`.
5. Updated client and publish call-sites to send/store full failed event payload.

---

## Files Changed (Path)

1. `db/schema.sql`
2. `utils/db/db-service.ts`
3. `pages/api/db/get-failed-publishes.ts`
4. `pages/api/db/track-failed-publish.ts`
5. `pages/api/db/clear-failed-publish.ts`
6. `utils/db/db-client.ts`
7. `utils/nostr/nostr-helper-functions.ts`

---
## Outcome / Impact

- Fixed `42P01` (`relation does not exist`) for failed publish endpoints.
- `GET /api/db/get-failed-publishes` now returns `200` reliably on fresh/local DBs.
- Retry flow is now more reliable because `event_data` is stored and can be replayed.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -128,8 +128,8 @@ CREATE INDEX IF NOT EXISTS idx_discount_codes_code ON discount_codes(code);
 -- Failed relay publish tracking table
 CREATE TABLE IF NOT EXISTS failed_relay_publishes (
     event_id TEXT PRIMARY KEY,
-    relays TEXT NOT NULL,
     event_data TEXT,
+    relays TEXT NOT NULL,
     created_at BIGINT NOT NULL,
     retry_count INTEGER DEFAULT 0
 );


calvadev (approved):
This was resolved in a separate PR, but still adds useful checks.

PR #231, opened 2026-02-19
feat: encrypted digital delivery & dynamic weight support

GautamBytes:
**Overview**
Introduces end-to-end encrypted digital asset delivery and adds support for dynamic weight-based pricing in product listings.

**Key Changes**
- Reliability: Added a Blossom server fallback loop. The uploader automatically tries multiple servers if a primary server rejects the encrypted binary.
- Weights: Added dynamic weight tiers (e.g., 100g, 1kg) with individual pricing logic. Total costs now update dynamically based on selection.
- UX:  Added an attachment preview card for sellers to verify or remove digital assets before publishing.
- Updated automated order DMs to explicitly include the selected weight for accurate fulfillment.
- Stability: Implemented Unicode-safe Base64 encoding to prevent application crashes caused by special characters in filenames.


utils/encryption/content-crypto.ts
calvadev, line 1: We can replace the AES encryption with NIP-44 for Nostr-native compatibility. You can use https://github.com/shopstr-eng/milk-market/blob/main/utils/encryption/file-encryption.ts as a reference.

PR #233, opened 2026-03-02
Site-wide performance optimization and checkout correctness

GautamBytes:
**This PR delivers a single-pass, performance-only optimization across app shell, data loading, route chunking, parsing, and caching, plus final checkout correctness fixes.**

**Key outcomes:**

- Reduced shared first-load JS from 394 kB baseline to 206 kB
- Reduced _app chunk from 835 kB raw baseline to 77.5 kB
- Route targets improved: /orders 197 kB, /listing 286 kB, /cart 266 kB
- Route-scoped bootstrap and lazy-loading of signer/nostr-heavy paths
- Startup/data fetch cleanup (less duplicate work, better staged loading)
- Route-level code splitting for heavy cart/listing/orders flows
- Parser/slug hot-path optimizations with bounded cache to avoid unbounded memory growth
- Polling reductions via event-driven storage updates
- PWA/service-worker tuning and middleware import hardening
- API/DB fetch scoping to reduce overfetching


components/home/marketplace.tsx
```diff
@@ -582,7 +592,7 @@ function MarketplacePage({
                     No reviews . . . yet!
                   </p>
                   <p className="mt-4 text-lg text-light-text dark:text-dark-text">
-                    Seems there aren&apos;t any reviews for this shop yet.
+                    Seems there aren&apos;t unknown reviews for this shop yet.
```
calvadev: Change back to `any` here

next.config.mjs
```diff
@@ -8,50 +8,53 @@ const withPWA = withPWAInit({
...
       },
     },
   ],
 });
 
 const nextConfig = {
-  bundlePagesRouterDependencies: true,
   output: "standalone",
   reactStrictMode: true,
-  turbopack: {},
```
calvadev, line 66: This needs to be added back, otherwise the project fails to build and run

GautamBytes: Ah my bad, reverted it

GautamBytes:
@calvadev this also seems ready now!

PR #260, opened 2026-04-02
fix: require signed Nostr event for MCP API key operations

jorisstrakeljahn:
### Description
The "signedEvent" verification on POST and DELETE in "pages/api/mcp/api-keys.ts" was wrapped in an optional if (signedEvent) block. This allowed requests without any signed event to skip signature verification entirely and create or revoke API keys for arbitrary pubkeys.

Changes:
- Make "signedEvent" mandatory on both POST and DELETE, returning "401" when missing
- Extract the repeated validation logic into a typed requireSignedEvent() helper to keep both handlers DRY  

### Resolved or fixed issue
none  

backend-only change, no UI impact

### Affirmation

### DIFF
--- a/components/settings/buyer-profile-form.tsx
+++ b/components/settings/buyer-profile-form.tsx
@@ -1,0 +1,199 @@
+import { useEffect, useState, useContext, useMemo } from "react";
+import { useRouter } from "next/router";
+import { useForm, Controller } from "react-hook-form";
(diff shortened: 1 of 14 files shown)


jorisstrakeljahn:
Thanks for the thorough follow-up! The replay protection with consumed event IDs and binding the proof to the specific operation are solid improvements, I hadn't considered those attack vectors.

Issue #261, opened 2026-04-02
[BUG] Product Listing Slug Collision Bug


Abhishek-Punhani:
### Description

I discovered a critical bug where multiple products with similar titles (that convert to the same URL slug) were being incorrectly routed to the wrong listing page. Specifically, when accessing a product listing with a simple slug that matches multiple products, the system consistently shows the first product in the array instead of the requested one.

### Current Behavior

When I clicked a product listing called "Last brick," it sometimes showed a completely different product (a van) with the same description and ID. However, when I opened the same link in a new tab (Ctrl+Click), it would sometimes show the correct product. This inconsistent behavior made the listing page unreliable and confusing.

### Expected Behavior

- Each product should display its own correct listing page, regardless of title similarities
- When multiple products have the same normalized slug, the system should require disambiguation using the seller's pubkey
- Simple slugs should only work for products without title collisions

### Steps to Reproduce

1. Open MarketPlace
2. Click on the listing named `Last Brick` with seller-npub: `npub1ndvrpzh4nunyj9zf5grc7r553s9ttjs62u77m42nxdd7987cxa4qg5nncw`.(There might be more entries, I just found a bug with this one).
3. Observe that the wrong product is displayed
4. Try opening the link in a new tab or refreshing - behavior may differ

### Next.js Version

16.1.7

### Operating System

Linux (Other)

### Node.js Version

v20.19.5

### Browser

Chrome

### Screenshots

When I opened the listing directly, it showed the wrong image, but opening it with ' ctrl+c ' showed the correct listing.
https://github.com/user-attachments/assets/27494aa6-87da-49db-8097-e4dabd150888

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #266, opened 2026-04-02
fix: improve slug matching logic to handle multiple exact matches

Abhishek-Punhani:
### Description
Improved slug matching logic to handle multiple exact matches by invoking pubkey-based slug matching.

### Resolved or fixed issue
#261 

### Steps to Reproduce
- Navigate to the marketplace and see `Last Brick` (Or other listing for which multiple exact entries exist)
- Verify the details are correct

### Affirmation
- [ ] My code follows the [CONTRIBUTING.md](https://github.com/hxrshxz/shopstr/blob/main/contributing.md) guidelines


GautamBytes:
The DB fallback now breaks valid exact slugs that happen to end in -xxxxxxxx.
Please keep exact slug matching first, and only use the pubkey-suffix path for real disambiguation cases.
The same issue also got introduced in the profile/shop slug helper.

PR #268, opened 2026-04-03
Resolves Client Side ProductContext hydration issue on listing page

ayushshrivastv:
SSR changes only, independent from relay.

### DIFF
--- a/components/__tests__/listing-page.test.tsx
+++ b/components/__tests__/listing-page.test.tsx
@@ -1,0 +1,220 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import { useRouter } from "next/router";
+import { nip19 } from "nostr-tools";
+import ListingPage, {
+  getServerSideProps,
+} from "../../pages/listing/[[...productId]]";
+import { ProductContext } from "@/utils/context/context";
+import { NostrContext } from "@/components/utility-components/nostr-context-provider";
+import {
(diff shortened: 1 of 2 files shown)


wolgwang1729 (commented):
LGTM on direction, but not ready to merge yet. I checked out this PR branch and validated the new listing-page tests locally.

- Targeted test result: 1 suite, 3 tests passed, 0 failed.
- The new SSR hydration coverage and stale-state reset scenario are passing smoothly.

Before this gets merged, I have a couple of quick follow-ups:

- **Blocking issue**: naddr links that include relay hints can fail the client-side identifier match, which clears the SSR-seeded listing state after hydration. This can cause a direct listing load to briefly render and then disappear. Please match naddr identities by decoding and comparing identifier/pubkey/kind fields instead of comparing encoded strings.
- **Test gap**: please add a case for an naddr URL with relay hints to ensure the SSR-seeded listing remains rendered after hydration. The current tests do not cover this variant.
- Kindly follow the [PULL_REQUEST_TEMPLATE.md](https://github.com/shopstr-eng/shopstr/blob/main/.github/PULL_REQUEST_TEMPLATE.md) as it requires affirming that the [CONTRIBUTING.md](https://www.google.com/search?q=https://github.com/shopstr-eng/shopstr/blob/main/contributing.md) guidelines were followed.

PR #269, opened 2026-04-03
Added a relay backed fallback for cache misses

ayushshrivastv:
The issue is split between [pages/listing](https://github.com/shopstr-eng/shopstr/blob/main/pages/listing/%5B%5B...productId%5D%5D.tsx) and [utils/nostr/fetch-service.ts](https://github.com/shopstr-eng/shopstr/blob/main/utils/nostr/fetch-service.ts). Right now if a direct listing URL is not already available in the local DB/cache or hydrated product context, the page can still fail because there is no targeted relay lookup path for that listing identifier.

This change adds a relay-backed fallback for direct listing loads, so when the listing is missing from local cache/context, the page can fetch it directly from relays using the listing identifier, cache the fetched event, and then render it. It also keeps the stale-state reset behavior so old listing data does not block a new direct route from resolving correctly.

### DIFF
--- a/components/__tests__/listing-page.test.tsx
+++ b/components/__tests__/listing-page.test.tsx
@@ -1,0 +1,204 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import { useRouter } from "next/router";
+import ListingPage from "../../pages/listing/[[...productId]]";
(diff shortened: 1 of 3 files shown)


ayushshrivastv:
Nip 50 has resolved this issue.

PR #271, opened 2026-04-03
fix search with Shopstr fetch/cache hydration

ayushshrivastv:
Shopstr currently relies on client-side filtering for marketplace search, which limits discoverability and makes search results depend too much on local state.

These changes improve marketplace discoverability while preserving existing wallet, Cashu, and checkout functionality.

The implementation builds on Shopstr's existing relay, parsing, caching, and UI entry points, allowing search to land as an incremental, protocol-focused change.


ayushshrivastv:
Closing this out — this was addressed and merged as part of the long term goal to add NIP 50 search support

PR #273, opened 2026-04-03
fix shopstr reporting with fetch/cache flow

ayushshrivastv:
Shopstr currently lacks a standardised reporting mechanism for objectionable listings and seller profiles.

These changes improve moderation and marketplace safety without affecting wallet, Cashu, or checkout functionality.

The implementation builds on Shopstr's existing relay, parsing, publishing, caching, and UI entry points, allowing reporting support to land as an incremental, protocol-focused change.

Screenshot 2026-04-03 at 8 55 39 AM

Screenshot 2026-04-03 at 8 57 55 AM

Screenshot 2026-04-03 at 8 58 00 AM

Screenshot 2026-04-03 at 8 58 17 AM


ayushshrivastv:
the report gets published first as a nostr kind 1984 event, and then it also gets cached locally in shopstr’s database in the report_events table so it can be fetched back through the existing data flow.

arnavkirti (commented):
A couple of small suggestions: 
- colocate `REPORT_TYPES` and `REPORT_TYPE_LABELS` to keep them in sync
- tighten the return type of `onSubmit` to `Promise<void>` or a specific result
- Optionally, decide whether empty details should be allowed and enforce that in the UI. 

Otherwise, LGTM!

GautamBytes:
This is heading in a good direction, but there are still 2 things to fix before it’s merge-ready.

Please add `report_events to deleteCachedEventsByIds()` so deleted reports don’t stay stuck in cache, and make `/api/db/fetch-reports` filter server-side instead of returning the full cached report set.

GautamBytes (approved):
LGTM

calvadev (approved):

PR #276, opened 2026-04-03
Make database optional and prevent crashes when DATABASE_URL is missing

23bsm038-Mahi:
### Problem

Previously, the app crashed when `DATABASE_URL` was not configured.
Database-related functions threw errors, causing API routes to return 500 responses and the UI to break.

### Solution

* Made database usage optional
* Updated DB initialization to avoid throwing errors when `DATABASE_URL` is missing
* Added fallback handling in all database-related functions
* API routes now return safe default responses instead of failing

### Result

* App no longer crashes without database configuration
* API routes return safe fallback data instead of 500 errors
* UI loads successfully even when DB is not set
* Improved developer experience for local development

### Summary

This change ensures the application remains stable and usable even when a database is not configured.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -1,0 +1,1 @@
+DATABASE_URL=postgres://postgres:postgres@localhost:5432/shopstr
--- a/pages/api/cache-events.ts
+++ b/pages/api/cache-events.ts
@@ -1,0 +1,2 @@
+export { config } from "./db/cache-events";
+export { default } from "./db/cache-events";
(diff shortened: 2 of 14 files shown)


23bsm038-Mahi:
hi @GautamBytes ,
Thanks for pointing this out!

I checked PR #233 — it mainly focuses on performance optimizations (bundle size, lazy loading, caching, etc.), while this PR specifically handles crashes when `DATABASE_URL` is missing and makes database usage optional with safe fallbacks.

So this addresses a different issue (stability rather than performance).

Please let me know if you'd like me to align anything with the existing changes !

PR #281, opened 2026-04-03
Fix NIP-05 verification fetch errors

### LINKED ISSUE #262: [BUG]  Fix NIP-05 verification failures caused by client-side CORS policy blocks
##  Bug Description
While testing the platform locally and inspecting the network logs, I noticed that client-side NIP-05 user identity verification requests are failing for several external domains (such as `primal.net`, `cheshirepleb.net`, and `proofofink.art`). 

Because the app (both local host and deployed webiste) attempts to fetch the `.well-known/nostr.json` file directly from these external servers, web browsers block the request due to missing `Access-Control-Allow-Origin` headers. This causes profile verification to fail silently

### Current Behavior

The browser's developer console displays red Failed to fetch or CORS policy errors when attempting to verify external user profiles. Because the browser intercepts and blocks the payload, the identity check fails are failing
 

### Expected Behavior

The application successfully retrieves the NIP-05 verification payload [NIP-05] and server api returns "verified:true"
for that user

### Steps to Reproduce

1. Run the application locally on `localhost:5000` or https://shopstr.market/.
2. Open your browser's Developer Tools and go to the **Console** or **Network** tab.
3. Observe the `Failed to fetch` or `CORS policy` red errors in the console.

### Next.js Version

16.1.7

### Operating System

Linux (Ubuntu)

### Node.js Version

24.14.1

### Browser

Chrome

### Screenshots

Image

### Additional Context

To solve this, instead of making the fetch call directly from the browser, we could route the request through the backend server . This bypasses the browser's CORS restrictions and successfully retrieves the .well-known/nostr.json payload which then can be sent to the frontend

### Code of Conduct

arnavkirti:
### Description
Fix NIP-05 verification so the client no longer fetches external `/.well-known/nostr.json` endpoints directly.
--
## Root cause
Profile verification was running in the browser and fetching external NIP-05 documents from third-party domains. Many of those domains do not send `Access-Control-Allow-Origin`, so browsers blocked the requests and valid verifications failed silently.

## Changes
- add a Next.js API route that performs the external NIP-05 fetch server-side with timeout handling
- update the client-side verifier to call the internal API route instead of the third-party domain directly
- add route-level tests covering success, malformed identifiers, and external fetch failures

## Impact
NIP-05 verification now avoids client-side CORS failures for external domains while preserving the existing boolean verification flow used during profile hydration.

### Resolved or fixed issue
Fixes #262 

### Affirmation

### For more details on what to include, see:

[Bug Report Template](https://github.com/shopstr-eng/shopstr/blob/main/.github/ISSUE_TEMPLATE/bug_report.md)
[Feature Request Template](https://github.com/shopstr-eng/shopstr/blob/main/.github/ISSUE_TEMPLATE/feature_request.md)
[Documentation Issue Template](https://github.com/shopstr-eng/shopstr/blob/main/.github/ISSUE_TEMPLATE/documentation_improvement.md)
[Security Issue Template](https://github.com/shopstr-eng/shopstr/blob/main/.github/ISSUE_TEMPLATE/security_vulnerability.md)


GautamBytes:
LGTM Now!!

utils/nostr/nostr-helper-functions.ts
```diff
@@ -1568,36 +1568,23 @@ export async function verifyNip05Identifier(
...
-      // Use a timeout to prevent hanging requests
-      const controller = new AbortController();
-      const timeoutId = setTimeout(() => controller.abort(), 10000); // 10 second timeout
-
-      const response = await fetch(url, { signal: controller.signal });
-      clearTimeout(timeoutId);
+      const response = await fetch(
+        `/api/nostr/verify-nip05?nip05=${encodeURIComponent(
+          nip05
+        )}&pubkey=${encodeURIComponent(pubkey)}`
```
Aryan0699: - The url is still relative , a absolute url is requied in Node 
- Can Remove @-split validation — server already handles this so no need to duplicate
- Can add type unknown as it would be more concrete 
- no need of nested try catch all three blocks were doing the same thing , one wrapper would be fine
I have suggested the optimum code below do give a check

```
export async function verifyNip05Identifier(
  nip05: string,
  pubkey: string,
  options?: { baseUrl?: string }
): Promise<boolean> {
  if (!nip05 || !pubkey) return false;

  try {
    const params = new URLSearchParams({ nip05, pubkey });
    const path = `/api/nostr/verify-nip05?${params.toString()}`;

    const baseUrl =
      options?.baseUrl ??
      (typeof window !== "undefined" ? window.location.origin : null);

    if (!baseUrl && typeof window === "undefined") {
      throw new Error("verifyNip05Identifier requires baseUrl in SSR");
    }

    const requestUrl = baseUrl
export async function verifyNip05Identifier(
  nip05: string,
  pubkey: string,
  options?: { baseUrl?: string }
): Promise<boolean> {
  if (!nip05 || !pubkey) return false;

  try {
    const params = new URLSearchParams({ nip05, pubkey });
    const path = `/api/nostr/verify-nip05?${params.toString()}`;

    const baseUrl =
      options?.baseUrl ??
      (typeof window !== "undefined" ? window.location.origin : null);

    if (!baseUrl && typeof window === "undefined") {
      throw new Error("verifyNip05Identifier requires baseUrl in SSR");
    }

    const requestUrl = baseUrl
      ? new URL(path, baseUrl).toString()
      : path;

    const response = await fetch(requestUrl);

    if (!response.ok) return false;

    const data: unknown = await response.json();

    return (
      typeof data === "object" &&
      data !== null &&
      "verified" in data &&
      (data as { verified?: boolean }).verified === true
    );
  } catch {
    return fals
      ? new URL(path, baseUrl).toString()
      : path;

    const response = await fetch(requestUrl);

    if (!response.ok) return false;

    const data: unknown = await response.json();

    return (
      typeof data === "object" &&
      data !== null &&
      "verified" in data &&
      (data as { verified?: boolean }).verified === true
    );
  } catch {
    return false;
  }
}
```

Aryan0699: These are few things that can be improved . I hope it helps  :thumbsup:

arnavkirti: Thank you so much for the review! I have addressed it in the latest commit.

arnavkirti:
@calvadev Please take a look at this PR!

calvadev (approved):

PR #282, opened 2026-04-03
Escape special characters in marketplace search queries

23bsm038-Mahi:
## Summary

Search currently treats user input as a regular expression, which causes unexpected behavior for queries containing special characters such as "c++", "foo.bar", or "[test]".

## Changes

- Trimmed the search query to remove extra whitespace
- Escaped regex metacharacters before constructing the search matcher
- Ensured user input is treated as a literal string instead of a regex pattern
- Added a regression test to verify that queries like "c++" return correct results

## Why this matters

This improves search reliability and user experience by ensuring common queries with special characters behave as expected.

## Scope

- Minimal and focused change
- Does not affect existing search logic beyond fixing incorrect matches
- No new dependencies introduced

## Testing

- Added test case for special character queries (e.g. "c++")
- Verified locally using:
  npm test -- --runInBand components/__tests__/display-products.test.tsx

### DIFF
--- a/components/__tests__/display-products.test.tsx
+++ b/components/__tests__/display-products.test.tsx
@@ -1,0 +1,96 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import DisplayProducts from "../display-products";
+import {
+  FollowsContext,
+  ProductContext,
+  ProfileMapContext,
+} from "@/utils/context/context";
+import {
+  NostrContext,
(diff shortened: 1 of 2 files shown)


calvadev (approved):

wolgwang1729:
@23bsm038-Mahi Why did you use `kind: 30018` instead of `kind: 30402`?

PR #291, opened 2026-04-03
Escape special characters in category filters

23bsm038-Mahi:
## Summary

Escape category values before constructing the regex used in category filtering.

## Problem

Category filters currently pass selectedCategory directly into RegExp. If a category contains special regex characters like "(", "[" or "+", it can throw an "Invalid regular expression" error and break filtering.

## Changes

- Escaped selectedCategory using the existing escapeRegExp() helper before creating the regex

## Why this matters

Prevents runtime errors caused by user-generated category names and improves marketplace reliability.

## Scope

- One-line change
- No changes to filtering behavior
- No new dependencies
- Safe and backward compatible

### DIFF
--- a/components/__tests__/display-products.test.tsx
+++ b/components/__tests__/display-products.test.tsx
@@ -1,0 +1,96 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import DisplayProducts from "../display-products";
+import {
+  FollowsContext,
+  ProductContext,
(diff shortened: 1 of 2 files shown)


23bsm038-Mahi:
Hi @calvadev @GautamBytes !
This ensures category filters handle special characters safely when categories come from user-generated data. Happy to add a test if needed!

PR #300, opened 2026-04-03
Control Trust toggle state from marketplace filter

23bsm038-Mahi:
## Summary

Make the Trust toggle a controlled component by binding it to the existing wotFilter state.

## Problem

The Trust filter switch updates the state on click, but does not receive the current state as a prop. This can cause the toggle UI to become out of sync with the actual filter state after re-renders or parent state updates.

## Changes

- Passed `isSelected={wotFilter}` to the Switch component in `components/utility-components/shopstr-switch.tsx`

## Why this matters

Ensures the toggle UI always reflects the actual filter state, improving reliability and user experience.

## Scope

- One-line UI-only change
- No changes to filtering logic
- Fully backward compatible

## Testing

- Verified that toggle state stays consistent after interactions and re-renders

### DIFF
--- a/components/__tests__/display-products.test.tsx
+++ b/components/__tests__/display-products.test.tsx
@@ -1,0 +1,120 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import DisplayProducts from "../display-products";
+import {
+  FollowsContext,
+  ProductContext,
+  ProfileMapContext,
+} from "@/utils/context/context";
+import {
+  NostrContext,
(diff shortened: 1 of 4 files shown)


kaihere14:
Nice catch! The switch was basically uncontrolled, so if it was missing the isSelected= {wotFilter} prop, it could visually drift from its actual state after a re-rendering occurs. Just to be sure; does the shopstr-switch.tsx component use isSelected or does it have a different name for that prop?

PR #330, opened 2026-04-04
feat: implement safe JSON parsing for localStorage with validation and error handling

arnavkirti:
### Description
- Introduce `parseJsonWithFallback` and `getLocalStorageJson` in `utils/safe-json.ts` to centralize typed, safe JSON parsing for `localStorage`, with optional validation and automatic key cleanup via `removeOnError`.

- Migrate cart, discount, and Nostr `localStorage` usage (including `relays`, `mints`, `tokens`, `history`, `signer`, etc.) to these helpers so malformed or invalid JSON no longer crashes rendering and instead recovers via validators and safe defaults.

- Broaden selected `localStorage`-backed types where needed while still validating shapes at the boundary, and add unit tests covering valid/malformed JSON, validator failures, key removal, and non-throwing recovery paths.

### Resolved or fixed issue
none

### Affirmation

### For more details on what to include, see:

[Bug Report Template](https://github.com/shopstr-eng/shopstr/blob/main/.github/ISSUE_TEMPLATE/bug_report.md)
[Feature Request Template](https://github.com/shopstr-eng/shopstr/blob/main/.github/ISSUE_TEMPLATE/feature_request.md)
[Documentation Issue Template](https://github.com/shopstr-eng/shopstr/blob/main/.github/ISSUE_TEMPLATE/documentation_improvement.md)
[Security Issue Template](https://github.com/shopstr-eng/shopstr/blob/main/.github/ISSUE_TEMPLATE/security_vulnerability.md)

### DIFF
--- a/components/utility-components/checkout-card.tsx
+++ b/components/utility-components/checkout-card.tsx
@@ -50,6 +50,24 @@ import { getLocalStorageJson } from "@/utils/safe-json";
 const SUMMARY_CHARACTER_LIMIT = 100;
 type CartDiscountsMap = Record<string, { code: string; percentage: number }>;
 
+const isCartDiscountsMap = (value: unknown): value is CartDiscountsMap => {
+  if (!value || typeof value !== "object" || Array.isArray(value)) {
+    return false;
+  }
+
+  return Object.values(value).every((entry) => {
+    if (!entry || typeof entry !== "object" || Array.isArray(entry)) {
+      return false;
+    }
+
+    const candidate = entry as { code?: unknown; percentage?: unknown };
+    return (
+      typeof candidate.code === "string" &&
+      typeof candidate.percentage === "number"
+    );
+  });
+};
+
 export default function CheckoutCard({
   productData,
   setInvoiceIsPaid,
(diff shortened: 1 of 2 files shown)


kaihere14 (commented):
Solid fix for a real crash vector raw JSON.parse on localStorage without a try/catch is a ticking bomb. The 
  abstraction makes sense.                                                                                      
                                                                                                                
  A few things:                                                                                                 
                  
  - removeOnError silently deletes the key on any parse failure or validation mismatch. For a cart or discount  
  key, that's destructive the user loses their data with no warning. At minimum log the error; ideally only
  remove on known corruption, not validator mismatches (which could be a schema version mismatch).              
  - parseJsonWithFallback swallows all errors and returns the fallback. If the caller passes a wrong fallback
  type (e.g. [] when the data is an object), it silently proceeds with bad state. The generic typing helps but  
  doesn't fully prevent this at runtime.
  - The nostr-helper-functions.ts change is a large diff for a "safe JSON" PR worth confirming that the signer
   type refactor is actually needed here or if it's scope creep.                                                
  - getLocalStorageJson does an SSR guard (typeof window === "undefined") but returns the fallback silently.
  Fine for now, but if this ever gets called server-side intentionally, it'll be hard to debug.                 
                  
  Tests are good. LGTM with the removeOnError behavior reconsidered.

PR #359, opened 2026-04-06
fix: update marketplace stats fetch to include timestamp and adjust d…

wolgwang1729:
### Description
This PR fixes the homepage stats first-load inconsistency in the Shopstr by the Numbers section (locally verified).

- Updated stats fetching in `pages/index.tsx` so it re-runs when product data length changes, instead of only once on initial mount.
- Added a cache-busting query param to the stats request to avoid stale cached responses returning outdated values.
- Local impact:
  - Completely fixes the first-load issue where the first two stat cards showed `0` or placeholders until manual refresh.

### Resolved or fixed issue
Fixes #358

### Video 

https://github.com/user-attachments/assets/bf453dfc-2393-42b3-9aab-7a4a2a495768

### Affirmation


wolgwang1729:
I am getting

## Error Type
Runtime SyntaxError

## Error Message
Unexpected end of JSON input


    at JSON.parse (<anonymous>:null:null)
    at <unknown> (utils/nostr/fetch-service.ts:378:38)

## Code Frame
  376 |             if (pubkeyProfilesToFetch.includes(event.pubkey)) {
  377 |               try {
> 378 |                 const content = JSON.parse(event.content);
      |                                      ^
  379 |                 const profile = {
  380 |                   pubkey: event.pubkey,
  381 |                   created_at: event.created_at,

Next.js version: 16.1.7 (Turbopack)

utils/nostr/fetch-service.ts
```diff
@@ -494,23 +519,30 @@ export const fetchProfile = async (
...
+            continue;
+          }
+
+          const profile: NipProfile = {
+            pubkey: event.pubkey,
+            created_at: event.created_at,
+            content,
+            nip05Verified: false,
+          };
+          if (content.nip05) {
```
wolgwang1729: Hi @arnavkirti, I rebased and resolved the conflicts in the branch. I kept the `fetch-service.ts`changes from PR #294, but I’d appreciate it if you could verify that the latest merge didn’t break any of the profile-handling behavior or the DB caching changes from your PR.

GautamBytes (approved):
LGTM!

wolgwang1729:
TODO: 

1. Currently, when the page loads for the first time, it shows only 50 sellers because only they have been inserted into the database at that point. This happens because the data is being loaded from the database while the background sync is still processing; after a manual reload, the correct (higher) number of sellers shows up. This needs to be handled so the first load reflects the full count.
2. There is a similar issue with images and products in the marketplace display. They fail to load on the first page visit and only appear after a second load or manual refresh.

calvadev:
Have the TODOs been addressed? @wolgwang1729

wolgwang1729:
@calvadev Not yet. I will be free most probably after SoB to address them.

GautamBytes (approved):

PR #367, opened 2026-04-07
Require auth for order status and message read routes

ayushshrivastv:
Unauthenticated order status update API enables tampering

https://github.com/shopstr-eng/shopstr/blob/7e26694692c065aa7e87b40bd8f91466830be60c/pages/api/db/update-order-status.ts#L1-L35

still accepts orderId, status, and messageId with no auth check, while 

https://github.com/shopstr-eng/shopstr/blob/7e26694692c065aa7e87b40bd8f91466830be60c/pages/api/db/mark-messages-read.ts#L1-L20

still accepts arbitrary messageIds with no auth check. the underlying db mutations in https://github.com/shopstr-eng/shopstr/blob/main/utils/db/db-service.ts#L856-L918 also still update message_events by id or order_id only, with no ownership validation


utils/db/db-service.ts
```diff
@@ -912,6 +912,65 @@ export async function getUnreadMessageCount(pubkey: string): Promise<number> {
...
+      if (buyerPubkey && sellerPubkey) {
+        break;
+      }
+    }
+
+    return { buyerPubkey, sellerPubkey };
+  } catch (error) {
+    console.error("Failed to get order participants:", error);
+    return { buyerPubkey: null, sellerPubkey: null };
+  } finally {
```
Aryan0699, line 965: If the DB query throws an error, this returns null, null. Then the route handler catches that in the update-order-status file and return 404 . Maybe we should let this error propagate instead of returning { null, null } here, so DB failures will return 500 instead of 404 as it would be catched by the outer catch .
Because now it will create confusion whether the order doesn't exist or the database is down. 
We can just replace the return with "throw error" so a DB failure now correctly returns 500, and the 404 path is only hit when the order genuinely doesn't exist

utils/db/db-service.ts
```diff
@@ -912,6 +912,65 @@ export async function getUnreadMessageCount(pubkey: string): Promise<number> {
...
+  const dbPool = getDbPool();
+  let client;
+
+  try {
+    client = await dbPool.connect();
+    const result = await client.query<{ tags: string[][] }>(
+      `SELECT tags
+       FROM message_events
+       WHERE order_id = $1 OR ($2 IS NOT NULL AND id = $2)
+       ORDER BY created_at DESC`,
```
Aryan0699, line 929: Try consider adding a LIMIT (5 or 10) here as no need to fetch all the rows since the required info of buyer seller pubkey are mostly present in the initial few messages only .

Aryan0699:
@ayushshrivastv would be great if you address the review comments when you get a chance.

ayushshrivastv:
>  A buyer can cancel but not ship. A seller manages the confirmation and pending is shared which makes sense as any party might need to set the intial state. Just a few small changes are suggested !!

that is the intended behavior here, and this pr already enforces that same role split. buyer can cancel but not ship, seller manages confirmed, shipped, and completed, and pending is shared since either side may need to set the initial state.

https://github.com/ayushshrivastv/shopstr/blob/66ee961713e405a8e94ad2cb1afc934123446449/pages/api/db/update-order-status.ts#L8-L33

Aryan0699:
Yes, that part looks correct 👍 I was just referencing it. The pending review comments I mentioned are about the other suggestions I added above.

ayushshrivastv:
> Yes, that part looks correct 👍 I was just referencing it. The pending review comments I mentioned are about the other suggestions I added above.

Can you elaborate?

Aryan0699:
image

These reviews present just above

Aryan0699:
Aren't they visible ?

ayushshrivastv:
> Aren't they visible ?

Actually, it’s not visible on my side!

Aryan0699 (changes requested):
The fix looks perfect . A buyer can cancel but not ship. A seller manages fulfillmen and pending is shared which makes sense as the initial state either party might need to set.
Just a small thing ,

ayushshrivastv:
OK, now it's visible.

Aryan0699:
Perfect thanks :+1:  Might be some glitch !!

GautamBytes:
can u solve the conflicts?

ayushshrivastv:
> can u solve the conflicts?

Ready for your review, @GautamBytes.

GautamBytes:
- `updateOrderStatus()` can still rewrite order_id on any message the caller participates in, so please either stop mutating order_id there or verify that the passed messageId already belongs to that same order before updating. 

- Second, since this PR is specifically securing POST mutation routes, please include a NIP-98 payload hash for the request body and validate it server-side too, so the auth proof is bound to the exact body being authorized.”

ayushshrivastv:
> * `updateOrderStatus()` can still rewrite order_id on any message the caller participates in, so please either stop mutating order_id there or verify that the passed messageId already belongs to that same order before updating.
> * Second, since this PR is specifically securing POST mutation routes, please include a NIP-98 payload hash for the request body and validate it server-side too, so the auth proof is bound to the exact body being authorized.”

Addressed! @GautamBytes 

Stopped messageId updates from rewriting `order_id` and tied NIP-98 auth to the exact POST body being sent.

GautamBytes:
LGTM Now!!

calvadev (approved):

PR #369, opened 2026-04-08
fix: user profile image upload issue

wolgwang1729:
### Description
- Fixes the profile image upload flow to prevent UI reversion and runtime edge-case failures after upload.
- Hardens form input handling for numeric fields by safely converting optional values to string using `value?.toString() || ""` in both profile form implementations.
- Improves uploader reliability by only setting loading state when files actually exist and by filtering out falsy upload results before invoking upload callbacks.
- Prevents accidental form submission during file upload by setting the uploader button to `type="button"`.
- Replaces NextUI `Input` with native `input` for file selection to avoid controlled file input conflicts.
- Adds Blossom response normalization to support both top-level JSON fields and `nip94_event.tags` responses.
- Adds a fallback URL constructor from SHA-256 hash when URL is omitted by a Blossom server and throws a clear error if no usable URL can be resolved.
- Prevents stale context-driven profile resets in settings by guarding repeated form resets with `if (contextLoadedRef.current) return;`.

Previous Error:
error now

After Fix:

https://github.com/user-attachments/assets/2097f473-5d89-465f-af41-e1e7a5197dd5

### Affirmation

- [X] My code follows the [CONTRIBUTING.md](https://github.com/hxrshxz/shopstr/blob/main/contributing.md) guidelines


GautamBytes:
can u solve the conflict? 
ping me for review then

GautamBytes (dismissed):
`pages/settings/user-profile.tsx` still does DB-first hydration and writes back created_at: 0, and the new contextLoadedRef can block fresher profileContext updates after that. Please switch this page to the same timestamped recency/local-fallback logic as `components/settings/user-profile-form.tsx` so stale data can’t win after an image upload.

wolgwang1729:
Great catch! I initially held off on changing the hydration approach in `user-profile.tsx` because it felt a bit outside the original scope of this PR, but I completely agree it's needed to prevent stale data from overriding the context.

I've now refactored both `user-profile.tsx` and `user-profile-form.tsx` to use the same timestamped local-fallback logic. The shared fallback functions are now in `nostr-helper-functions.ts`, so both files use a single source of truth for profile hydration.

GautamBytes:
LGTM Now!!

calvadev (approved):

PR #389, opened 2026-04-09
pull listing search predicate into a standalone utility and add NIP-56 report helpers

tharu-jwd:
The filter functions in DisplayProducts had no unit tests and testing them meant spinning up the whole component with all its context providers. Moved them to utils/search/listing-filters.ts so they're plain functions that are easy to test in isolation.

Also added typed helpers in utils/nostr/nip56.ts for building kind 1984 report tags for profiles and listings since that's needed for the reporting flow anyway.

29 tests, all passing.

### DIFF
--- a/components/display-products.tsx
+++ b/components/display-products.tsx
@@ -1,6 +1,6 @@
 import { useState, useEffect, useContext } from "react";
-import { nip19 } from "nostr-tools";
 import { deleteEvent } from "@/utils/nostr/nostr-helper-functions";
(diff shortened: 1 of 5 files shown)


GautamBytes:
the new `utils/nostr/__tests__/nip56.test.ts` file is failing npm run type-check because tags[0] is treated as possibly undefined in a few assertions. If you make those test assertions type-safe and rerun type-check, this should be good to merge

GautamBytes:
LGTM Now!!

PR #390, opened 2026-04-10
Fix MCP onboarding quick-start correctness

YASH-ai-bit:
## Summary

This updates MCP onboarding so the generated connection details and quick-start examples match the current MCP HTTP flow more closely.

## What changed

- Updated [pages/api/mcp/onboard.ts](/c:/Users/gujja/SOB/shopstr/pages/api/mcp/onboard.ts) to resolve the onboarding base URL more safely:
  - prefer `NEXT_PUBLIC_BASE_URL` when configured
  - otherwise use `x-forwarded-proto` / `x-forwarded-host` when present
  - otherwise fall back to the local request protocol instead of hardcoding `https`

- Updated the generated onboarding quick-start examples to:
  - include `Accept: application/json, text/event-stream`
  - use `curl -i` for `initialize` so the `Mcp-Session-Id` response header is visible
  - add clearer guidance about reusing the returned session id

- Added focused regression coverage in [pages/api/mcp/__tests__/onboard.test.ts](/c:/Users/gujja/SOB/shopstr/pages/api/mcp/__tests__/onboard.test.ts) for:
  - local HTTP onboarding URLs
  - `NEXT_PUBLIC_BASE_URL` override with trailing-slash normalization
  - forwarded host/protocol handling
 
- Documented MCP onboarding setup requirements in [env.example](/c:/Users/gujja/SOB/shopstr/env.example), including `MCP_ENCRYPTION_KEY` and the optional `NEXT_PUBLIC_BASE_URL`

## Why

While testing MCP onboarding locally, the response was generating `https://localhost:5000/...` URLs even when the local server was running over plain HTTP. The quick-start examples also did not include the `Accept` header required by the current MCP endpoint, which made first-run manual testing more confusing than it needed to be.

This keeps the onboarding flow the same, but makes the generated response more accurate for local development and proxied deployments.

## Testing

- `npm test -- pages/api/mcp/__tests__/onboard.test.ts`
- `npm test -- utils/mcp/__tests__/auth.test.ts utils/mcp/__tests__/metrics.test.ts utils/mcp/__tests__/request-proof-server.test.ts`

## Notes

Manual smoke testing confirmed:
- onboarding returns the expected local `http://localhost:5000/...` MCP URLs
- MCP initialize still succeeds and returns a valid `Mcp-Session-Id`
- existing MCP auth / metrics / request-proof tests still pass

### DIFF
--- a/__tests__/pages/api/mcp/onboard.test.ts
+++ b/__tests__/pages/api/mcp/onboard.test.ts
@@ -47,10 +47,23 @@ const FIXED_NSEC = "nsec1testonboard";
 const FIXED_NPUB = "npub1testonboard";
 const FIXED_AGENT_NAME = "Test Agent";
 
-type MockResponse = NextApiResponse & {
+type MockResponse = {
   statusCode: number;
   jsonBody: unknown;
+  status(code: number): MockResponse;
+  json(payload: unknown): MockResponse;
 };
+
+function setNodeEnv(value: "development" | "production" | "test" | undefined) {
+  const env = process.env as Record<string, string | undefined>;
+
+  if (value === undefined) {
+    delete env["NODE_ENV"];
+    return;
+  }
+
+  env["NODE_ENV"] = value;
+}
 
 function createMockRequest(
   overrides: Partial<NextApiRequest> = {}
@@ -72,7 +85,7 @@ function createMockRequest(
 }
 
 function createMockResponse(): MockResponse {
-  const res = {
+  const res: MockResponse = {
     statusCode: 200,
     jsonBody: undefined,
     status(code: number) {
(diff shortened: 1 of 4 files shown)


GautamBytes (changes requested):
A few things to address before merge:

- Test ships as a production route. next build exposes /api/mcp/__tests__/onboard.test. Move the test outside pages/ or configure pageExtensions in next.config.mjs to exclude *.test.ts.

- Type-check fails. The MockResponse cast at line 75 doesn't satisfy NextApiResponse. Use Partial<NextApiResponse> or a minimal stub type instead.

- Agent names break curl examples. An apostrophe in the name (e.g. O'Reilly Bot) escapes the -d '...' wrapper and can inject shell commands on copy-paste. Escape or sanitize the name before embedding.

- Forwarded headers trusted without validation. x-forwarded-host/x-forwarded-proto are used unchecked, letting an attacker redirect onboarding URLs + curl examples to their domain. Restrict proto to http/https, and require NEXT_PUBLIC_BASE_URL in production or validate the host against an allowlist.

Happy to review once these are addressed!

YASH-ai-bit:
hey @GautamBytes, I have made changes, can you review them please. Thank you!

GautamBytes (approved):
LGTM

calvadev (approved):

PR #396, opened 2026-04-10
feat: implement address management components and functionality

Abhishek-Punhani:
Implemented address management components and functionality

### Description
- Add `AddressPicker` component for selecting and managing saved addresses.
- Enhance PreferencesPage to include address management features.
- Added `Use Saved Address` in cart view to use saved address during checkout.
- Add unit tests for saved address helper functions.
- Update local storage helper functions to manage saved addresses.
- Define SavedAddress type in typescript types.

### Resolved or fixed issue
Fixes #374 

https://github.com/user-attachments/assets/559b55f3-561a-4579-8028-190b20e31342

### Affirmation

- [ ] My code follows the [CONTRIBUTING.md](https://github.com/hxrshxz/shopstr/blob/main/contributing.md) guidelines


arnavkirti (commented):
Thanks for the solid work on this PR - Would like to point out some things:
- Please confirm that we validate all required fields (`name`, `address`, `city`, `state`, `zip`, `country`) both on the client and at the boundary where these are sent to the backend.
- Consider what happens if `zip` or `country` are missing or malformed (especially for non‑US addresses) and add tests to cover these edge cases.
- `isDefault` is a boolean on `SavedAddress`, but there is no obvious guard to ensure only one address can be default at a time.
- Can you add logic (and tests) to ensure we never end up with multiple defaults, and clarify the behavior when the current default is deleted?
- Please double‑check that all interactive elements in the Address picker / management UI are keyboard accessible and have appropriate ARIA labels where needed.
- If we are showing destructive actions (like delete), a confirmation step or undo pattern would be helpful to prevent accidental loss of an address.
- it would be good to have a few focused tests on:
  - Rendering of saved addresses list.
  - Creating a new address and setting it as default.
  - Editing/deleting an existing address.
  - Ensuring only one default address after updates.

Abhishek-Punhani:
> Thanks for the solid work on this PR - Would like to point out some things:
> 
> * Please confirm that we validate all required fields (`name`, `address`, `city`, `state`, `zip`, `country`) both on the client and at the boundary where these are sent to the backend.
> * Consider what happens if `zip` or `country` are missing or malformed (especially for non‑US addresses) and add tests to cover these edge cases.
> * `isDefault` is a boolean on `SavedAddress`, but there is no obvious guard to ensure only one address can be default at a time.
> * Can you add logic (and tests) to ensure we never end up with multiple defaults, and clarify the behavior when the current default is deleted?
> * Please double‑check that all interactive elements in the Address picker / management UI are keyboard accessible and have appropriate ARIA labels where needed.
> * If we are showing destructive actions (like delete), a confirmation step or undo pattern would be helpful to prevent accidental loss of an address.
> * it would be good to have a few focused tests on:
>   
>   * Rendering of saved addresses list.
>   * Creating a new address and setting it as default.
>   * Editing/deleting an existing address.
>   * Ensuring only one default address after updates.


Ensuring one default address and destructive action modal is there already, could you once pull the changes and test it before suggesting

Abhishek-Punhani:
Handling the pin codes and duplicate checks , ARIA, these I will handle in another PR

GautamBytes (changes requested):
1. Can you fix the saved-address default handling. Right now updating/saving an existing default address can leave the address book with no default at all. We should always preserve exactly one default address.

2. Also, Please update the helper tests to use the real saved-address helpers with mocked `localStorage` instead of re-implementing the logic inside the test file. As written, the tests can still pass even when the actual helper behavior is wrong.

Once those are fixed, ping me again and I’ll re-review.

GautamBytes:
can u solve the conflicts?

GautamBytes (approved):
LGTM Now!!

calvadev:
A nice bow to tie this off would be to add a "save address" button or option in the checkout flow after inputtig the information to make that easier and not require prior settings navigation to input it

Abhishek-Punhani:
> A nice bow to tie this off would be to add a "save address" button or option in the checkout flow after inputtig the information to make that easier and not require prior settings navigation to input it

@calvadev, There is already a save option in the checkout flow via the checkbox “Save this address for future use” under Shipping Information, so users can save the entered address without needing to go to Settings first.

PR #402, opened 2026-04-12
 restore real-time subscriptions by migrating to subscribeMany

abhijit9040:
This PR fixes a critical regression in the Nostr subscription engine where a removed method subscribeMap was being called on the SimplePool instance.

Changes:- 
- Migrated legacy subscribeMap usage to the modern subscribeMany API in utils/nostr/nostr-manager.ts.
- Simplified the relay/filter mapping logic by removing unnecessary flattening of the requests object.
- Ensured compatibility with nostr-tools v2 architecture.

Testing Performed:-
- Ran npm run type-check to verify that the TS2551 error is resolved.
- Verified that the NostrSub object is correctly constructed with the appropriate relay URLs and filters.


calvadev:
subscribeMany is actually the deprecated method, subscribeMap is updated. You probably need to bump your nostr-tools version to get it working again.

PR #403, opened 2026-04-13
Fix : Product card dropdown click isolation on marketplace (prevent unintended listing navigation)

Aryan0699:
### Issue

- Clicking the seller dropdown area inside a marketplace product card was opening the listing page, instead of only opening/interacting with the dropdown because the card content was wrapped in Next Link although nested controls were present

### Description
For opening the dedicated dropdown only and not moving to listing page following things were changed

- Removed the Link wrapper from the card body and switched to controlled navigation handlers.
- Added a interaction guard to block navigation for nested interactions like carousel controls, dropdown menus, profile dropdown region .
- Added explicit click navigation handler for normal card clicks.
- Added auxiliary click handler to support middle-click new-tab behavior.
- Preserved parent override behavior: if onProductClick prevents default, card route push is skipped.

### Resolved or fixed issue
`none`  

### Before
[Screencast from 2026-04-13 15-49-54.webm](https://github.com/user-attachments/assets/854035ff-1675-44e2-a276-9b3bfaf53be2)

### After
image

### Affirmation

### DIFF
--- a/components/display-products.tsx
+++ b/components/display-products.tsx
@@ -250,7 +250,10 @@ const DisplayProducts = ({
     return `/listing/${product.id}`;
   };
 
-  const onProductClick = (product: ProductData, e?: React.MouseEvent) => {
+  const onProductClick = (
+    product: ProductData,
+    e?: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>
+  ) => {
     setFocusedProduct(product);
     if (product.pubkey === userPubkey) {
       e?.preventDefault();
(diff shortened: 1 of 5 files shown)


wolgwang1729 (commented):
Nice catch! But a few suggestions:
1. I can no longer right-click on the card and open it in a new tab.
2. Make the user know that clicking on the seller name opens a dropdown, maybe by changing that seller name component's behaviour when it is hovered. (The current cursor change behaviour isn't that eye-catching)

Aryan0699:
@wolgwang1729 Thanks for pointing that out . Done the requied changes :+1:

wolgwang1729:
LGTM

GautamBytes (approved):
LGTM Now!!

calvadev (approved):

PR #404, opened 2026-04-13
Fix navbar profile flickering on page load

wolgwang1729:
### Description
Fixed a flickering issue in the navbar where merchant profiles would temporarily show shop information (name and image) instead of user information during page reloads.

**Changes:**
- Modified `fetchCachedEvents` in `shopstr/utils/db/db-service.ts` to explicitly include the event `kind` in the SQL `WHERE` clause.
- Ensures that queries for user profiles (kind 0) do not return shop profiles (kind 30019) even when they share the same database table and pubkey.
- Fixed a bug where the database would return the most recent event from the `profile_events` table regardless of kind, causing incorrect data to be served to the `/api/db/fetch-profile` endpoint.

### Screenshots

Before:

https://github.com/user-attachments/assets/33254510-4d23-46cc-a42c-39c826f62eb1

After:

https://github.com/user-attachments/assets/266ef333-5df5-4e20-a52d-cf2d01a4c715

### Affirmation

### DIFF
--- a/utils/db/db-service.ts
+++ b/utils/db/db-service.ts
@@ -688,9 +688,9 @@ export async function fetchCachedEvents(
 
   try {
     client = await dbPool.connect();
-    let query = `SELECT id, pubkey, created_at, kind, tags, content, sig FROM ${table} WHERE 1=1`;
-    const params: any[] = [];
-    let paramIndex = 1;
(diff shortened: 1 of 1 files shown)


GautamBytes (approved):
LGTM

calvadev (approved):

PR #405, opened 2026-04-13
fix: use local time for expiration datetime inputs

Shreyas2004wagh:
## Description

Fix expiration `datetime-local` inputs to use local wall-clock time instead of UTC-formatted strings.

The current code was building input values and `min` constraints with `toISOString().slice(0, 16)`. That converts dates to UTC first, but `datetime-local` inputs are interpreted in the browser as local time. In non-UTC timezones, this can shift the displayed expiration time and make the `min` value incorrect.

## What this changes

- add a small shared helper for formatting `datetime-local` values from local date parts
- use the local formatter for product expiration defaults when editing a listing
- use the local formatter for product expiration `min`
- use the local formatter for discount code expiration `min`
- add focused tests for the datetime-local formatting helper

## Why this matters

This keeps the read/edit path consistent with the existing submit path, which already parses `datetime-local` input values as local time.

Without this fix:
- an existing expiration can appear shifted when editing
- the `min` constraint can be offset by timezone
- users near timezone boundaries can select values that are already in the past locally

## Testing

- `npm test -- datetime-local.test.ts --runInBand`

### DIFF
--- a/components/my-listings/discount-codes.tsx
+++ b/components/my-listings/discount-codes.tsx
@@ -9,6 +9,7 @@ import {
 } from "@nextui-org/react";
 import { TrashIcon } from "@heroicons/react/24/outline";
 import { SHOPSTRBUTTONCLASSNAMES } from "@/utils/STATIC-VARIABLES";
+import { formatCurrentDateTimeLocalValue } from "@/utils/datetime-local";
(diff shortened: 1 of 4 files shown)


GautamBytes (approved):
LGTM

calvadev (approved):

PR #406, opened 2026-04-14
fix: harden storefront cart state and re-resolve shop routes on navigation

Shreyas2004wagh:
## Description

Fix two storefront stability issues:

1. storefront cart badge polling could break when `localStorage.cart` contained malformed JSON
2. shop pages could keep stale resolution state when navigating between different storefront slugs

## What was the issue

The storefront wrappers were polling `localStorage.cart` and using raw `JSON.parse`. If the cart value was corrupted, the parse would throw and the storefront cart quantity logic could break repeatedly.

The shop route pages also kept local lookup state across slug changes. In practice, client-side navigation between different `/shop/{slug}` or `/shop/{slug}/{page}` routes could keep the old shop mounted or skip re-resolving the new slug.

## What this changes

- add a small helper for safely reading storefront cart quantity from local storage
- update storefront cart badge logic to use safe JSON parsing instead of raw `JSON.parse`
- reset shop route lookup state when the slug changes
- ignore stale async lookup results after route changes
- remove the one-way route resolution latch from `/shop/[...shopPath]`
- add regression tests for malformed cart storage and shop route re-resolution

## Why this matters

This makes storefront navigation more resilient:
- corrupted cart storage no longer breaks storefront cart badge updates
- navigating between storefronts no longer risks showing stale shop data
- late async responses for a previous slug no longer overwrite the current route state

## Testing

- `npm test -- storefront-cart.test.ts --runInBand`
- `npm test -- shop-route-resolution.test.tsx --runInBand`


GautamBytes (approved):
LGTM

calvadev (approved):

PR #223, opened 2026-02-12
Complete Website Revamp

GautamBytes:
COMPLETE WEBSITE REVAMP

image
-----
image
-----
image
-----
image
-----
image
-----
image
-----
image
-----
image

PR #408, opened 2026-04-15
fix contrast on user-type buttons by switching to a darker yellow variant for dark theme

theAnuragMishra:
### Description
White text on bright yellow background isn't very easy on eyes.
image

This pr adds a new yellow variant `shopstr-yellow-dark` set to be `#534e3c`.

Now it looks like this:
image

### Resolved or fixed issue
none

### Affirmation

### DIFF
--- a/pages/onboarding/user-type.tsx
+++ b/pages/onboarding/user-type.tsx
@@ -60,7 +60,7 @@ const UserTypeSelection = () => {
                 onClick={() => setSelectedType("buyer")}
                 className={`flex flex-1 flex-col items-center justify-center rounded-lg border-2 p-8 transition-all ${
                   selectedType === "buyer"
-                    ? "border-shopstr-purple bg-shopstr-yellow dark:border-shopstr-yellow dark:bg-shopstr-yellow"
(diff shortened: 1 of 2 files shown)


theAnuragMishra:
update: also removed the purple border around the yellow bg in light mode.

GautamBytes:
> e purple border around the yellow bg in light mode.

Hey @theAnuragMishra , I actually have a pending pr for website revamp might be a little trouble for u but can u checkout that and see if this problem persist with new design also? I guess it won't be there.

[Check here](https://github.com/shopstr-eng/shopstr/pull/223)

theAnuragMishra:
Hmmmm it looks like you have changed the onboarding flow itself. I can't see the current `user-type.tsx` component. It seems like you now offer a person to modify both profile types one by one (skip if they don't want) in the flow? @GautamBytes

GautamBytes:
Yeah i guess its a mistake from my end, would correct it later. Contrast issue is solved in revamped one but for now i think let's merge urs

GautamBytes (approved):
LGTM

PR #417, opened 2026-04-16
Escape storefront policy HTML rendering

ayushshrivastv:
**Security Issue**: policy.content is treated like markdown text, but inlineFormat() does not escape HTML before passing it into dangerouslySetInnerHTML. That means raw attacker HTML survives unchanged into a public storefront page. The same origin also exposes sensitive client side state in localStorage, including serialised signer objects and NIP 46 app keys.

The full exploitation process is demonstrated in the attached video:
 https://www.notion.so/Security-Vulnerabilities-3367d68f8c31800394d6d88b56674914?source=copy_link

Screenshot 2026-04-16 at 8 47 40 PM

Screenshot 2026-04-16 at 8 47 55 PM
Screenshot 2026-04-16 at 8 48 06 PM

So basically, a seller account was able to save a malicious JavaScript payload inside the storefront’s Terms of Service page. Then when another user (the victim) opened that public policy page, the script automatically ran in their browser session. Because of that, the payload could access and read data from the victim’s localStorage within their browser.

### DIFF
--- a/components/storefront/__tests__/storefront-policy-page.test.tsx
+++ b/components/storefront/__tests__/storefront-policy-page.test.tsx
@@ -1,0 +1,52 @@
+import { render, screen } from "@testing-library/react";
+import StorefrontPolicyPage from "@/components/storefront/storefront-policy-page";
+
+const colors = {
+  primary: "#000000",
+  secondary: "#ffffff",
+  accent: "#111111",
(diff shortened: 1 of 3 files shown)


GautamBytes (approved):
LGTM. Thanks:)

calvadev (approved):

PR #418, opened 2026-04-16
Unauthenticated cached event deletion API allows cross user data removal

ayushshrivastv:
Security Issue: The server trusts a browser supplied list of event IDs for a destructive action, but it does not verify that the caller is authorised to delete those events. Authorisation is missing both at the route level and at the database mutation level.

On current `main`, the issue is that Shopstr’s cached event deletion API trusts a client-supplied list of eventIds and deletes them from the database without verifying who is making the request or whether those events belong to that pubkey. The API route at [`pages/api/db/delete-events.ts`](https://github.com/shopstr-eng/shopstr/blob/main/pages/api/db/delete-events.ts#L1-L20) accepts eventIds from the request body and immediately calls deleteCachedEventsByIds(eventIds). Then in [`utils/db/db-service.ts`](https://github.com/shopstr-eng/shopstr/blob/main/utils/db/db-service.ts#L781-L840), the database logic deletes matching IDs across multiple cached event tables purely by id, including `product_events`, `review_events`, `message_events`, `profile_events`, `wallet_events`, `community_events`, and `config_events`. There is no ownership check there.

This is a missing server side authorisation check on a destructive endpoint. Even if the normal UI only deletes a user’s own events, that restriction is not enforced by the backend. A user can tamper with the request payload and submit arbitrary event IDs, and the server will trust them.

**Fix**

The fix should require a signed ownership proof on /api/db/delete-events and validate it server side before any deletion happens. The request should be bound to the exact eventIds being deleted, and the backend should verify that all referenced events belong to the authenticated pubkey before executing the mutation. The DB path should fail closed if ownership cannot be confirmed.


theAnuragMishra:
@calvadev this should be priority !!
it's a big security hole. i found it too and was going to work but saw this pr.

theAnuragMishra (commented):
Looks good overall!

calvadev (approved):

PR #422, opened 2026-04-17
fix: normalize cart invoice totals across product and shipping flows

Shreyas2004wagh:
## Description

Normalize cart invoice totals so checkout, order summary, and payment flows all use the same per-product totals.

## What was the issue

The cart page was building a totals map one way, while the invoice/payment flow was reading it in two different ways:

- the cart page stored base totals by seller pubkey
- order summary and shipping adjustments were reading totals by product id
- payment token splitting was also reading totals by seller pubkey

That made the data model inconsistent. In carts with multiple items from the same seller, earlier totals could be overwritten, and shipping-adjusted totals were being computed in temporary objects without becoming the actual source of truth for payment.

## What this changes

- store base cart totals per product id in the cart page
- add cart total helpers for shipping-adjusted per-product totals and final total summation
- make checkout keep a current per-product totals map for the active shipping mode
- use the current per-product totals for order summary line items and payment token amounts
- add focused tests for the cart totals helpers

## Why this matters

This keeps checkout totals consistent across:
- cart line items
- shipping mode selection
- order summary persistence
- payment and donation amount calculation

It also fixes the overwrite risk for multi-product carts from the same seller.

## Testing

- `npm test -- cart-totals.test.ts --runInBand`
- `npx tsc --noEmit --ignoreDeprecations 6.0`


GautamBytes (approved):
LGTM

PR #423, opened 2026-04-17
Hide the created api key by default with a button to toggle visibility

theAnuragMishra:
### Description
When the api key is created, it is visible by default which is not the standard way I see on most websites. It should be hidden by default with a button to toggle visibility. Most of the time a user is gonna just copy it without seeing it anyway.

Screenshot 2026-04-17 at 19 57 40

### Affirmation


GautamBytes (approved):
LGTM!!

calvadev (approved):

PR #424, opened 2026-04-17
fix: guard chat npub decoding against malformed content

Shreyas2004wagh:
## Summary
- guard buyer pubkey extraction in chat messages when npub-like text is malformed
- fall back to an empty buyer pubkey instead of letting nip19.decode throw
- add a regression test for malformed npub content

## Testing
- npm test -- chat-message.test.tsx --runInBand


GautamBytes (approved):
LGTM NOW!!

calvadev (approved):

PR #430, opened 2026-04-18
Preserve required product field compatibility

### LINKED ISSUE #416: [FEATURE] refactor: unify product event parsing with shared base parser and consumer adapters (UI and MCP)
### Description

Product parsing logic currently exists in two places:
- `parseTags()` for UI
- `parseProductEvent()` for MCP tools

Both functions parse the same NIP-99 product events but return different shapes and support different tag subsets. This creates duplication and increases the risk of parser drift when new product tags are introduced or existing logic is updated.

It would be better to have a shared base parser as the single source of truth, with consumer-specific adapters for the UI, MCP, and future integrations.

\cc @calvadev @GautamBytes 

### Feature Description

Introduce a unified product parsing architecture with:
**Shared Base Parser**
A canonical parser responsible for reading raw product events and parsing all supported product tags into a normalized internal ProductData shape.

**Consumer Adapters**
Separate adapters that transform canonical parsed data into consumer-specific output:
- UI Adapter → preserves UI state and frontend-friendly structures
 - MCP Adapter → JSON-safe output with pricing blocks and flattened structures

### Problem This Feature Solves

Current duplication causes several maintainability issues:
- New tags must be added in multiple places
- Bug fixes must be duplicated manually
- UI and MCP responses can silently diverge
- Some tags may be supported in one parser but missing in another
- Testing parsing logic becomes harder

### Proposed Solution

Refactor current product parsing into two layers:
- **Layer 1: Shared Parser**
- **Layer 2: Adapters**

### Alternative Solutions

_No response_

### User Impact

This improves Shopstr reliability for users by ensuring:
- Consistent product data across UI and MCP tools
- Faster rollout of new product tag support
- Fewer bugs caused by mismatched parsers

### Relevant NIPs

_No response_

### Implementation Ideas (Optional)

_No response_

### Screenshots or Mockups

_No response_

### Additional Context

_No response_

### Code of Conduct

Abhishek-Punhani:
### Summary

PR #612 superseded the original shared-parser implementation. This PR now contains only the compatibility behavior that remained useful:

- Maps `required_customer_info` to legacy UI `required` when `required` is missing.
- Maps legacy `required` to MCP `requiredCustomerInfo` when `required_customer_info` is missing.
- Preserves distinct values when both tags are explicitly present.
- Adds regression tests for all three cases

### Related issue

Follow-up compatibility fix for #416 after #612.


wolgwang1729 (commented):
LGTM. 
(You didn't tick the acknowledgement, though)

GautamBytes (approved):
LGTM

Abhishek-Punhani:
Hey @GautamBytes , was just looking back at this, should I rebase this for merging or its resolved already and for other prs as well?

GautamBytes:
Requested a review from calva on this and also pinging @Aryan0699 to look into below possible todos to fully close 416 issue.

- Extract a truly shared parser
Move the canonical parsing into a package-neutral module that both the Next app and packages/shopstr-mcp can import. It should not import React/UI code.

- Unify standalone MCP
Replace packages/shopstr-mcp/src/parse-tags.ts product parsing with the shared canonical parser + MCP adapter.

- Preserve standalone MCP fields
Make sure the shared model supports the richer MCP fields already present there: priceStatus, stock, visibility, productType, productFormat, structured images, shippingOptions.

- Add regression tests
Add tests proving root MCP, standalone MCP, and UI all parse the same event consistently, with adapter-specific shapes.

- Tighten invalid price behavior
Avoid silently converting malformed prices to 0; represent missing/invalid price distinctly for MCP, and let UI filtering hide invalid listings.

GautamBytes:
@Aryan0699 have a look at above msg whenever feasible!

Aryan0699:
Sure . I totally missed this. Will have a look at it👍

Aryan0699:
Hi @Gautam, the PR seems to correctly unify the root MCP and the UI parser. However the standalone MCP package needs to be wired with it.
The above plan seems correct, just there is one thing I would like to clarify. Since the MCP package is standalone and builds independently, introducing a shared parser would introduce a link between them and MCP would not be separate. Below are some options in which it can be handled:

- We can keep a duplicate of the file inside our package but that is what we are trying to solve.
- As mentioned, we can extract it into a third tiny package and both the root Next.js app and MCP server will depend on it. But this would add unnecessary complexity for a few parser files.
- The best approach I think would be to keep the shared file in utils/parsers/ only and let the standalone package include it at build time. We can have a prebuild npm script that copies the file into the MCP package before every tsc build. This way the standalone package stays independent and there wont be any drift between the canonical source and the copy.

GautamBytes:
https://github.com/shopstr-eng/shopstr/pull/543#issuecomment-4750045457
Check comment there. Proceeding as discussed

GautamBytes (approved):

PR #432, opened 2026-04-18
Unauthenticated delete events API lets any browser user delete shared cached listings by ID

ayushshrivastv:
The delete events API accepted arbitrary event id lists from any browser client and deleted shared cached event rows without verifying a signed request or ownership. The marketplace read path then treated that cache as authoritative for product batches and listing SSR, so a DevTools POST could remove another seller's cached listing from public responses until the event was re-cached from relays.

This change requires a signed Nostr request proof on /api/db/delete-events, signs the client-side delete helper with the existing request-auth pattern, and resolves the targeted cached rows before deletion so only events owned by the signing pubkey can be removed. It also adds a focused regression test that covers unsigned deletes, foreign-owner deletes, and the normalized happy path.

DevTools exploitation path

1. Open the live Shopstr site in a browser and open DevTools Console.
2. Enumerate public cached listing IDs:

```
const products = await fetch('/api/db/fetch-products?limit=5').then(r => r.json());
const victimId = products[0].id;
victimId
```
3. Delete the shared cached listing directly, without signing any deletion proof:

```
await fetch('/api/db/delete-events', {
method: 'POST',
headers: { 'Content-Type': 'application/json' },
body: JSON.stringify({ eventIds: [victimId] })
}).then(r => r.json());
```
4. Confirm the server-side mutation:

```
const after = await fetch('/api/db/fetch-products?limit=5').then(r => r.json());
after.some(p => p.id === victimId) // false
```

5. Because listing SSR calls `fetchProductByIdFromDb(identifier)`, opening `/listing/<victimId>` now loses the cached server side listing record until it is re-cached from relays.

This crosses a security boundary because the attacker is not just editing their own DOM or local state. They are issuing an unauthenticated request that mutates centralised server data used by other users and by server-rendered pages.

The code explicitly relies on frontend behaviour for the security decision as the frontend generates a signed Nostr deletion event But the backend never checks that such a deletion was actually authorised before deleting from the database.


calvadev:
There seems to be no changes here. Should this be closed or renamed?

ayushshrivastv:
> There seems to be no changes here. Should this be closed or renamed?

Currently working on this, holding off on the final commit until I can confirm it fully resolves the issue.

PR #439, opened 2026-04-19
Harden product image URLs and referrer policy

ayushshrivastv:
Shopstr accepts arbitrary product image tags from NIP-99 listing events and then renders those URLs in product/listing flows. A seller-controlled image URL can therefore still be used as a privacy leak, because the browser will request that third-party host and expose request metadata such as IP address, user agent, and timing.

After maintainer discussion, this PR is scoped to a lighter first-step hardening rather than a full centralized image proxy. It now:

- adds a site-wide `Referrer-Policy: no-referrer` header
- normalizes product image URLs in the product and zapsnag parsers
- keeps local static image paths working
- rejects dangerous schemes and malformed image values
- rejects local `/api/...` image targets
- rejects localhost, `.localhost`, `.local`, private IPv4 ranges, IPv6 loopback/link-local/ULA, protocol-relative URLs, and IPv4-mapped IPv6 loopback/private-style inputs

This does not fully remove third-party image host metadata exposure for arbitrary public image hosts; it intentionally avoids introducing server-side proxy load and CDN dependence, matching the lighter approach discussed in the PR thread.

Focused coverage lives in `utils/__tests__/images.test.ts`, `utils/parsers/__tests__/product-parser-functions.test.ts`, and `utils/parsers/__tests__/zapsnag-parser.test.ts`.


theAnuragMishra:
im not exactly aware of the image flow. so are the images processed on shopstr servers  or do the clients fetch the image from the provided url directly?

> blocks localhost

??

ayushshrivastv:
> im not exactly aware of the image flow. so are the images processed on shopstr servers or do the clients fetch the image from the provided url directly?  blocks localhost

the browser was fetching product images directly from the seller provided URL. After this change, product images go through the Shopstr \/api/product-image` proxy instead.

the fetch is being done by Shopstr through /api/product-image. In that case, localhost means the Shopstr server’s localhost, not the visitor’s. If that server side proxy did not block localhost or private IPs, an attacker could try to make Shopstr fetch internal services, admin ports, metadata endpoints, or internal API paths from the server environment itself. That is an SSRF risk.

theAnuragMishra:
Have you thought about the performance consideration? Or the load this gonna put on the server? 
If the browser uses `img` for showing images then even malicious image urls are harmless.

ayushshrivastv:
> Have you thought about the performance consideration? Or the load this gonna put on the server? If the browser uses `img` for showing images then even malicious image urls are harmless.

I did think about that tradeoff. The concern here is not that \` would execute JavaScript directly, but that on current main the browser is making a request to an attacker controlled image URL. That still gives the seller or image host a direct metadata leakage path for every viewer, like IP, user agent, timing, and listing interest.

For the proxy side, yes, it does add server load, so I tried to keep that bounded. The route is rate limited, has a max image size cap, only allows `http/https`, blocks `localhost/private` targets, rejects internal `/api/` paths, and sets cache headers. So this is meant to reduce the privacy/security issue without turning it into an unbounded fetch path.

But the optimal long term solution is a bounded cached proxy for untrusted remote product images, while letting trusted first party or approved host media be served directly. That keeps the privacy boundary on Shopstr side without making this an unbounded server fetch path.

calvadev:
Yeah performance here can take a hit at a tradeoff that isn't really super necessary. There isn't really any security vulnerabilities, mainly privacy related ones that are present in virtually all internt services unless they otherwise claim to not allow for any IP leakage.

And before images are uploaded (at least from Shopstr's side), they're stripped of all metadata and reconstructed.

We would also require a native CDN and introduce centralized image proxying that wouldn't allow the marketplace to load images if the server is down.

calvadev:
From Claude:

Lighter alternatives that cover ~80% of the threat:

1. <meta name="referrer" content="no-referrer"> — one line, kills the Referer leak. Free.
2. Referrer-Policy: no-referrer header — same thing, server side.
3. CSP img-src allowlist — restrict images to a known set of hosts (nostr.build, imgur, etc.). Sellers using random hosts get a broken image instead of a tracker. This is what most Nostr clients do.
4. Recommend/require Blossom or nostr.build-style content-addressed image hosts at upload time. Doesn't fix existing listings but channels new ones to safe hosts.
5. Lazy-load + warn on first hover — UX nudge so the user opts into loading off-domain media.

You could combine 1 + 3 and get most of the privacy benefit at ~zero infrastructure cost.

calvadev:
I could see a benefit for implementing this as a NSFW filter and mask those images/filter them out entirely, but that would be a separate feature.

ayushshrivastv:
> Yeah performance here can take a hit at a tradeoff that isn't really super necessary. There isn't really any security vulnerabilities, mainly privacy related ones that are present in virtually all internt services unless they otherwise claim to not allow for any IP leakage.
> 
> And before images are uploaded (at least from Shopstr's side), they're stripped of all metadata and reconstructed.
> 
> We would also require a native CDN and introduce centralized image proxying that wouldn't allow the marketplace to load images if the server is down.

Ya, I was treating this as a privacy / metadata leakage concern rather than a classic direct security bug. The reason I raised it was that a seller controlled image URL can still force direct browser requests to a third party host and leak viewer IP / request metadata.

But I agree full image proxying has real tradeoffs: more server load, more centralisation, CDN dependence, and images failing if Shopstr infra is down. So I’m okay treating the full proxy model as too heavy for this issue and scoping the fix down to basic URL hardening only, if that fits better with Shopstr’s design goals.

calvadev:
> > Yeah performance here can take a hit at a tradeoff that isn't really super necessary. There isn't really any security vulnerabilities, mainly privacy related ones that are present in virtually all internt services unless they otherwise claim to not allow for any IP leakage.
> > And before images are uploaded (at least from Shopstr's side), they're stripped of all metadata and reconstructed.
> > We would also require a native CDN and introduce centralized image proxying that wouldn't allow the marketplace to load images if the server is down.
> 
> Ya, I was treating this as a privacy / metadata leakage concern rather than a classic direct security bug. The reason I raised it was that a seller controlled image URL can still force direct browser requests to a third party host and leak viewer IP / request metadata.
> 
> But I agree full image proxying has real tradeoffs: more server load, more centralisation, CDN dependence, and images failing if Shopstr infra is down. So I’m okay treating the full proxy model as too heavy for this issue and scoping the fix down to basic URL hardening only, if that fits better with Shopstr’s design goals.

I think that would be a safer first-step approach.

theAnuragMishra:
> I could see a benefit for implementing this as a NSFW filter and mask those images/filter them out entirely, but that would be a separate feature.

@calvadev that is a better candidate for a client side feature?

ayushshrivastv:
After the changes I did here "Commit [7ba1544](https://github.com/shopstr-eng/shopstr/pull/439/commits/7ba154441043bc0a8aa9c70bc3e2c931788a10d6)", one thing I want to mention is that without proxying or host restriction, arbitrary public third party image hosts still receive viewer IP and still can request metadata. So this works as a lighter privacy step, but it does not fully remove that remaining direct `third party` tracking path.

> I could see a benefit for implementing this as a NSFW filter and mask those images/filter them out entirely, but that would be a separate feature.

Also, if we use an NSFW filter, the browser still loads the remote image in order to decide whether to blur, classify, or preview it, then the request already happened and the third party host still gets the viewer metadata, so the privacy leak is already there before the NSFW filter helps.

GautamBytes:
@calvadev pushed some changes,  see once if changes makes sense now to merge or need some tweak!

GautamBytes:
@calvadev can you review it once whenever feasible!

ayushshrivastv:
.

GautamBytes (approved):

PR #449, opened 2026-04-28
Validate listing mint quotes server-side

ayushshrivastv:
Switched the listing checkout path to ask a Shopstr API route for the quote, with the route fetching the listing and recomputing the amount before calling the mint.

Issue: Currently frontend loads mint metadata, creates a mint quote directly against mint.minibits.cash, then polls that quote. Shopstr’s backend is not in the path to recompute and lock the expected listing amount before invoice creation.

`the listing total is computed client side as ts, then passed through e2() into e3(s, i), where e is used directly in createMintQuoteBolt11(e). The only pre invoice validation is the client side e1(e) check, which verifies the amount is greater than zero and required fields are present. `

As now, I can simply modify ts before it reaches e2()

on this file https://shopstr.store/_next/static/chunks/0x-4m6wgrbb2_.js

The listing price must be treated as untrusted input from the client. The Shopstr backend needs a validation endpoint that independently resolves the correct price before minting any invoice. 

image

image

image


GautamBytes (approved):
LGTM Now!!

Aryan0699 (commented):
Hi Arnav, Just a few improvements that can be done

- Currently in verify-payment.ts the mint URL comes from a map and it is safe since it would always be the trusted mint URL, but it would be better to validate it against the trusted mints list before using it, just to guarantee that it is always valid regardless of how that value got there.
- Also in the Cashu token path (create-order.ts), when we extract the mint URL from the decoded token payload, that URL is still fully controlled by the buyer since they control the token they are submitting and the mint URL is embedded inside it. That URL could point to their fake server, and since we are passing it directly into CashuMint() without any check against the trusted mints list, the server will make an outbound request to it - which is an SSRF vulnerability.

GautamBytes:
@ayushshrivastv can you resolve conflicts, ping me once after that!

ayushshrivastv:
Conflicts are resolved and pushed in a03ba15. @GautamBytes

GautamBytes (approved):
LGTM!!

GautamBytes:
A small conflict here also @ayushshrivastv

ayushshrivastv:
> A small conflict here also @ayushshrivastv

Resolved!

PR #454, opened 2026-04-29
fix: cancel stale search requests on new query to prevent race condition in marketplace

### LINKED ISSUE #453: [BUG] Marketplace search missing request cancellation — stale results can overwrite current query
### Description

The marketplace search debounces correctly (via the useDebounce hook) but does not cancel in-flight work when a new query fires before the previous one completes. This means stale results from a slower earlier query can overwrite the correct results from a faster later query — a classic search race condition.

### Current Behavior

When a user types a search query, the debounce fires after 300ms and triggers a search pass. If the user continues typing before results return and a second debounced search fires, both passes are now in flight simultaneously. Whichever one completes last sets the displayed results — regardless of which query the user actually intended. A slow stale result silently overwrites a correct current result with no error or warning.

### Expected Behavior

When a new search query fires, any in-flight search work from the previous query should be cancelled immediately. Only the results from the most recent query should ever update the UI. This is the standard behaviour users expect from any search input.

### Steps to Reproduce

1. Open the marketplace page
2. Type a search query quickly, character by character (e.g. "bitcoin hardware")
3. Before results stabilize, continue typing to change the query (e.g. add " wallet")
4. Observe that results may flicker or settle on results from an intermediate query rather than the final one
5. The problem is more visible on slower connections or when the product list is large, since filter passes take longer

To reproduce deterministically in code: wrap the search useEffect with an artificial setTimeout delay and fire two rapid query changes — the second result will be overwritten by the first.

### Next.js Version

16.2.3

### Operating System

macOS

### Node.js Version

22.4.0

### Browser

Chrome

### Screenshots

_No response_

### Additional Context

_No response_

### Code of Conduct

nayan9617:
### Description

This change prevents stale marketplace product loads from writing outdated results into the UI when the app initialization flow reruns or the user navigates away mid-fetch.

**What was fixed:**
- Adds an `AbortController` to the app-level initialization flow so the previous product load is cancelled when a newer run starts or the component unmounts.
- Makes the product loader abort-aware across the DB batch fetch and relay fetch stages with proper cleanup.
- Updates the Nostr relay fetch helper to handle abort signals and close subscriptions safely without unhandled rejections.
- Centralizes abort short-circuit checks in `fetchAllPosts` to avoid duplication and reduce future drift.
- Improves abort-error detection to work robustly across different runtime environments.

**How it works:**
1. When the app initializes products, it creates an `AbortController` and passes its signal through the entire load pipeline.
2. If the component unmounts or a new initialization run starts, the old `AbortController` is aborted.
3. The product loader checks the abort signal at key checkpoints (before DB batches, after DB batches, before relay fetch, after relay fetch, during relay event merge).
4. If aborted, the loader cleanly resolves with any partial results and doesn't write to context.
5. Relay subscriptions are closed asynchronously and errors are caught to prevent unhandled rejections.

**Testing:**
- `fetchAllPosts` stops cleanly when aborted before the first DB batch resolves (no stale context writes).
- The abort signal propagates through to the relay fetch stage and closes subscriptions on abort.
- All existing tests continue to pass.
 
 
 ### Resolved #453


GautamBytes (changes requested):
Two things to address:
1. `NostrManager.fetch` now catches any `subscribe` failure and resolves `[]`. That can hide real relay/config errors as “no products found”. It should only resolve empty on abort; other errors should reject/rethrow.
2. `npm run lint` currently fails due to unused catch bindings in `nostr-manager.ts`.

nayan9617:
I'll scope the silent resolution strictly to abort, rethrow everything else, and clean up the unused catch bindings. Will push shortly.

GautamBytes (approved):
LGTM Now!

wolgwang1729 (approved):

GautamBytes (approved):

PR #458, opened 2026-04-30
fix: hide trust toggle without user follows

GautamBytes:
## Summary
- Remove the default follow-list fallback from the Trust filter
- Hide the Trust toggle unless the user is signed in and has direct follows
- Use the latest kind 3 contact list event when building the trust graph

## Why
The Trust toggle should represent the signed-in user’s own Nostr follow graph. Previously, logged-out users or users with no follows could still see default filtered sellers, which made the Trust filter misleading.


GautamBytes:
cc: @calvadev

calvadev (approved):

GautamBytes:
@calvadev ready for merge

calvadev (approved):

PR #477, opened 2026-05-05
 fix: pin MCP sessions to originating API key and add TTL eviction

kaihere14:
## Description

- **Session hijack prevention:** After `initialize`, subsequent POST/GET/DELETE requests that resume a session via `Mcp-Session-Id` now verify that the Bearer token resolves to the same `apiKey.id` that created the session. If there is a mismatch, the request is rejected with a `403` and JSON-RPC error `"Session belongs to a different API key"`. Previously, any valid API key could attach to any existing session, allowing a leaked session ID to be hijacked by a different credential — the tool closures would still operate under the original user's identity and permissions.
- **Session TTL with automatic eviction:** Sessions now track `createdAt` and `lastActivityAt` timestamps. A background sweep runs every 5 minutes and removes sessions that have been inactive for more than 30 minutes, calling `transport.close()` before deletion. This prevents unbounded memory growth from clients that crash or disconnect without sending DELETE.
- **Activity tracking:** Each successful session resume (POST and GET) updates `lastActivityAt`, so active sessions are kept alive while idle ones are cleaned up.

All changes are in `pages/api/mcp/index.ts`.

## Resolved or fixed issue

none

## Screenshots (if applicable)

N/A — backend-only security fix, no UI changes.

## Affirmation

### DIFF
--- a/pages/api/mcp/index.ts
+++ b/pages/api/mcp/index.ts
@@ -28,10 +28,37 @@ async function ensureTables() {
   }
 }
 
-const sessions = new Map<
-  string,
-  { transport: StreamableHTTPServerTransport; apiKey: ApiKeyRecord }
->();
+const SESSION_TTL_MS = 30 * 60 * 1000;
+const SWEEP_INTERVAL_MS = 5 * 60 * 1000;
+
+interface McpSession {
+  transport: StreamableHTTPServerTransport;
+  apiKey: ApiKeyRecord;
+  createdAt: number;
+  lastActivityAt: number;
+}
+
+const sessions = new Map<string, McpSession>();
+
+function rejectSessionMismatch(res: NextApiResponse) {
+  return res.status(403).json({
+    jsonrpc: "2.0",
+    error: { code: -32000, message: "Session belongs to a different API key" },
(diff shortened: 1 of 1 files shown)


GautamBytes (approved):
Looks good overall, just one TTL edge case before merge.
Expired sessions can be revived before the sweep runs, since resume updates lastActivityAt immediately.
Can you reject/delete expired sessions on resume before refreshing activity?

GautamBytes (changes requested):
Looks good overall, just one TTL edge case before merge.
Expired sessions can be revived before the sweep runs, since resume updates lastActivityAt immediately.
Can you reject/delete expired sessions on resume before refreshing activity?

kaihere14:
> Looks good overall, just one TTL edge case before merge. Expired sessions can be revived before the sweep runs, since resume updates lastActivityAt immediately. Can you reject/delete expired sessions on resume before refreshing activity?

i fixed it i hope now there is no edge case thank you!

GautamBytes (approved):
LGTM Now!

calvadev (approved):

PR #502, opened 2026-05-29
feat: add NIP 50 search support for marketplace listings

ayushshrivastv:
this PR adds Nip 50 search for marketplace listings so Shopstr can ask search capable relays for matching products instead of only filtering locally. It keeps regular browsing separate, uses backup search relays, deduplicates listing results, and handles slow relays better so one bad relay does not break the whole search.

Screenshot 2026-06-10 at 04 58 45

Screenshot 2026-06-10 at 05 30 19

Screenshot 2026-06-10 at 05 16 47

### DIFF
--- a/components/__tests__/display-products.test.tsx
+++ b/components/__tests__/display-products.test.tsx
@@ -628,7 +628,9 @@ describe("DisplayProducts search filtering", () => {
 
     await waitFor(() => {
       expect(screen.getAllByText("Deduped Coffee")).toHaveLength(1);
-      expect(screen.queryByTestId("product-relay-product")).not.toBeInTheDocument();
+      expect(
+        screen.queryByTestId("product-relay-product")
+      ).not.toBeInTheDocument();
       expect(screen.getByTestId("product-local-product")).toBeInTheDocument();
     });
   });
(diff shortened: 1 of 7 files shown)


GautamBytes (approved):
LGTM!

calvadev:
There's one thing to add here: some relays don't actually support NIP-50 so there should be a default backup list for searching if user -selected ones don't work.

calvadev:
Some default search relays that I'm aware of are

- relay.noswhere.com
- search.nos.today
- antiprimal.net
- relay.ditto.pub

ayushshrivastv:
> There's one thing to add here: some relays don't actually support NIP-50 so there should be a default backup list for searching if user -selected ones don't work.

thanks for the list! i've added all four as fallback defaults relay.noswhere.com, search.nos.today, antiprimal.net, and relay.ditto.pub. so now if the user's relays don't support NIP 50, search will fall back to these automatically.

Screenshot 2026-06-10 at 05 00 55

Screenshot 2026-06-10 at 05 09 46

Screenshot 2026-06-10 at 05 09 00

GautamBytes (approved):
Looks good to merge now

PR #504, opened 2026-06-01
Validate OpenGraph preview URLs before rendering

ayushshrivastv:
there is a script injection issue on this line... 

https://github.com/shopstr-eng/shopstr/blob/e1f7dc33b8d75201f55d223941a3a18e6599699e/pages/api/og-preview.ts#L206 
https://github.com/shopstr-eng/shopstr/blob/e1f7dc33b8d75201f55d223941a3a18e6599699e/components/messages/link-preview.tsx#L73

When a user sends a URL in chat... Shopstr fetches the page's OpenGraph metadata to generate a link preview. The backend returns the og:url as is... no validation check. So if a seller controls that page... they can set og:url to javascript:... and when the buyer clicks the preview card... that script runs directly in their Shopstr browser context.

### DIFF
--- a/components/messages/__tests__/link-preview.test.tsx
+++ b/components/messages/__tests__/link-preview.test.tsx
@@ -1,0 +1,54 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import "@testing-library/jest-dom";
+
+import LinkPreview from "../link-preview";
+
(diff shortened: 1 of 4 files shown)


GautamBytes (approved):
LGTM!

PR #511, opened 2026-06-04
Avoid persisting Cashu proofs in localStorage

ayushshrivastv:
Shopstr saves decrypted Cashu wallet proofs directly into browser localStorage under tokens. Anyone with DevTools access can open Application storage or just run localStorage.getItem("tokens") and copy spendable wallet proofs... which can lead to fund theft.

https://github.com/shopstr-eng/shopstr/blob/e1f7dc33b8d75201f55d223941a3a18e6599699e/pages/_app.tsx#L786-L788

the core change is: stop writing spendable Cashu proofs to localStorage["tokens"] and instead hold them in a module level in memory variable, accessed via a `setCachedCashuProofs()` setter exported from `nostr-helper-functions.ts`.

Every `localStorage.setItem("tokens", JSON.stringify(...))` across the codebase across 9 components is replaced with `setCachedCashuProofs(...)`, legacy localStorage tokens are migrated, and transaction critical wallet paths no longer depend on `getLocalStorageData().tokens`. This avoids exposing spendable proofs through browser storage.


GautamBytes:
@calvadev can you review it once whenever feasible?

calvadev:
This is a good start! Been reviewing to make sure there aren't any regressions in functionality. Will be dropping a comment on what to revise.

calvadev:
The claim paths call publishProofEvent(...) before setCachedCashuProofs(...), with no try/catch. A mint quote can only be minted once, so if that publish throws on a relay error, the claim is skipped, the pending quote stays open, and the next boot re-attempts the mint -> the mint returns "already issued" -> the recovery state machine marks it failed_terminal. That's a real fund-loss window. It's made worse by the callers (product-invoice-card.tsx / cart-invoice-card.tsx) not wrapping the recovery call either, so the throw also skips markMintQuoteClaimed and the user-facing modal.

Maybe credit the in-memory cache first, then make the publish best-effort. Durability is already guaranteed by the encrypted retry queue, and setCachedCashuProofs de-dupes by proof secret.

There's also still a narrow window where if the relay publish and the encrypted-queue write both fail at the same instant, proofs live only in the cache until reload. We could close it later by having publishProofEvent return { published, queued } and only completing the claim once at least one durable path succeeded.

ayushshrivastv:
> The claim paths call publishProofEvent(...) before setCachedCashuProofs(...), with no try/catch. A mint quote can only be minted once, so if that publish throws on a relay error, the claim is skipped, the pending quote stays open, and the next boot re-attempts the mint -> the mint returns "already issued" -> the recovery state machine marks it failed_terminal. That's a real fund-loss window. It's made worse by the callers (product-invoice-card.tsx / cart-invoice-card.tsx) not wrapping the recovery call either, so the throw also skips markMintQuoteClaimed and the user-facing modal.
> 
> Maybe credit the in-memory cache first, then make the publish best-effort. Durability is already guaranteed by the encrypted retry queue, and setCachedCashuProofs de-dupes by proof secret.
> 
> There's also still a narrow window where if the relay publish and the encrypted-queue write both fail at the same instant, proofs live only in the cache until reload. We could close it later by having publishProofEvent return { published, queued } and only completing the claim once at least one durable path succeeded.

The mint claim path now credits the Cashu proof cache first, then marks the pending quote claimed, then does the Nostr publish as best effort. So a relay error won’t keep the quote open anymore, and the next boot won’t hit the “already issued” path and mark it `failed_terminal`. Product/cart recovery path goes through the same helper now, so publish failures don’t bubble up and skip `markMintQuoteClaimed` or the modal.

The encrypted retry queue still handles failed publishes, and the cache write uses the existing proof-secret de-dupe behavior.

The remaining “publish + queue write both fail” window is still a follow up. A cleaner fix there would be to make `publishProofEvent` return something like `{ published, queued }`, then only finalize once one durable path is confirmed.

GautamBytes:
@ayushshrivastv please solve the conflicts!

ayushshrivastv:
> @ayushshrivastv please solve the conflicts!

Resolved!

GautamBytes:
@calvadev can u review it once?

PR #521, opened 2026-06-06
Potential fix for code scanning alert no. 40: Clear text storage of sensitive information

calvadev:
Potential fix for [https://github.com/shopstr-eng/shopstr/security/code-scanning/40](https://github.com/shopstr-eng/shopstr/security/code-scanning/40)

General fix: do not store sensitive proof material unencrypted in `localStorage`. Encrypt before persistence and decrypt only when needed. In browser code, a practical approach is Web Crypto (AES-GCM) with a key derived from a user-controlled passphrase (PBKDF2). This preserves functionality (persist/reload tokens) while removing plaintext-at-rest.

Best single fix in the shown code:
- In `utils/cashu/wallet-recovery.ts`, replace direct `JSON.stringify(proofArray)` storage with encrypted payload storage.
- Add helper methods in the same file to:
  - derive/import an AES key from a passphrase,
  - encrypt JSON text with random IV and salt,
  - serialize encrypted blob for storage.
- Read passphrase from a runtime value (for example `process.env.NEXT_PUBLIC_WALLET_ENCRYPTION_KEY` fallback guarded for browser usage). If unavailable, fail closed (do not write plaintext).
- Keep `history` behavior unchanged unless it also contains secrets (it currently does not).

Test updates in `utils/cashu/__tests__/wallet-recovery.test.ts`:
- Mock `crypto.subtle`/`getRandomValues` behavior or mock encryption helper (if exported) so tests remain deterministic.
- Update assertions to verify `"tokens"` is no longer plain array JSON and that writing still occurs.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/utils/cashu/__tests__/wallet-recovery.test.ts
+++ b/utils/cashu/__tests__/wallet-recovery.test.ts
@@ -25,6 +25,7 @@ const mkProof = (secret: string, amount = 10): Proof =>
 
 describe("recoverProofsToBuyerWallet", () => {
   beforeEach(() => {
+    process.env.NEXT_PUBLIC_WALLET_ENCRYPTION_KEY = "test-wallet-key";
     window.localStorage.clear();
     helpers.getLocalStorageData.mockReset();
     helpers.publishProofEvent.mockReset();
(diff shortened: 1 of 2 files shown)


calvadev:
Alternative PR approach preferred.

PR #529, opened 2026-06-09
Tighten NIP-99 listing tests

ayushshrivastv:
This tightens the NIP-99 listing tests so the coverage now proves the behavior Shopstr depends on instead of relying on looser product fixtures.

The parser tests now use kind 30402 directly and cover a full classified listing shape, including content as the description, d tags, images, categories, price, shipping, quantity, and raw event handoff. I also added coverage for relay-hinted naddr decoding and the abortable fetch path that merges cached and relay listings by pubkey plus d tag.


GautamBytes (approved):
LGTM!

PR #534, opened 2026-06-10
Validate cart mint quotes server-side

GautamBytes:
## Summary

This PR closes the remaining payment validation gaps around cart checkout and MCP order creation.

- Adds a server-side cart mint quote endpoint that fetches the latest listing events, validates cart item selections, validates discount codes, recomputes totals, and creates the Cashu mint quote using the server-controlled trusted mint.
- Updates cart Lightning and NWC checkout to use the server-returned quote amount and product totals before polling, minting proofs, and splitting seller/donation/change tokens.
- Fixes MCP `create-order` so non-sat listing totals are converted to sats before creating Lightning quotes or checking Cashu token amounts.
- Adds shared helpers for Cashu-safe sat amounts, currency conversion, and trusted mint selection.

## Note

The new `/api/cart/mint-quote` route treats the request body as untrusted runtime input. It validates cart order type, pickup preference, item shape, selected variants, discount-code map shape, listing availability, stock, expiration, and quantity before calling the mint.

This PR intentionally does not change direct browser Cashu-token checkout; that path does not create a mint quote. It should still get a separate server-validated amount/totals follow-up, but it is separate from the Lightning/NWC quote-manipulation issue.

## Coordination

This complements #449, which validates single-listing Lightning/NWC mint quotes server-side. So ideally should merge after that.

### DIFF
--- a/__tests__/pages/api/cart/mint-quote.test.ts
+++ b/__tests__/pages/api/cart/mint-quote.test.ts
@@ -1,0 +1,290 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+
+const fetchProductByIdFromDbMock = jest.fn();
+const fetchProductByDTagAndPubkeyMock = jest.fn();
+const fetchAllProfilesFromDbMock = jest.fn();
+const validateDiscountCodeMock = jest.fn();
+const createMintQuoteBolt11Mock = jest.fn();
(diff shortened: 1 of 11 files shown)


GautamBytes:
@calvadev ready for review!

PR #543, opened 2026-06-17
Implement Read Tools - List Companies , Get Company Details, Get Storefront, Get Seller Reputation 

Aryan0699:
### Description

- Adds relay-backed MCP seller/profile tools: get_company_details, get_storefront, list_companies and get_seller_reputation
- Registered and exported all 5 new tools so they are visible to the MCP client.
- get_company_details:  Fetches and deduplicates both kind: 0 (user profile) and kind: 30019 (shop profile), alongside the seller's products and reviews.
- get_storefront:  Returns shop configuration and products but omits the heavy customer reviews.
- list_companies: Discovers public Shopstr sellers (querying kind: 30019). Add until pagination support .
- get_seller_reputation: Calculates and normalizes seller trust scores using kind: 31555 review tags (50% thumb weight, 50% category averages) - Based on Gamma Market Spec. 
- Extract review helpers and common seller utilities into seperate modules
- Extract duplicated guard pattern
- Add deduplication for profiles 
- Add  test coverage for the newly added tools (some are left)

### Video
https://drive.google.com/file/d/1xot-BLDSwV1DnTd5aNkZiZh0OV40QfhC/view?usp=sharing
### Affirmation

### DIFF
--- a/packages/shopstr-mcp/README.md
+++ b/packages/shopstr-mcp/README.md
@@ -3,9 +3,9 @@
 Standalone read-only MCP server package for Shopstr marketplace data.
 
 This package currently contains the standalone MCP shell, shared read-only
-infrastructure, and the first relay-backed read tools for public Shopstr
-marketplace data. Seller, storefront, reputation, prompt, and resource features
-will be added in follow-up PRs.
+infrastructure, relay-backed product/review tools, and relay-backed
+seller/storefront/reputation tools for public Shopstr marketplace data. Prompt
+and resource features will be added in follow-up PRs.
 
 ## Current Scope
 
@@ -13,8 +13,10 @@ will be added in follow-up PRs.
 - Reads relay, timeout, cache, and log-level settings from environment
   variables.
 - Starts an MCP server over stdio for local MCP-compatible clients.
-- Registers relay-backed core read tools:
-  `search_products`, `get_product_details`, and `get_reviews`.
+- Registers relay-backed read tools:
+  `search_products`, `get_product_details`, `get_reviews`,
+  `list_companies`, `get_company_details`, `get_storefront`,
+  and `get_seller_reputation`.
 - Registers disabled resource and prompt placeholders so `resources/list` and
   `prompts/list` return valid empty lists until those features are added.
 - Provides reusable infrastructure modules for upcoming tools:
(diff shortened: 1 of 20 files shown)


Aryan0699:
Will resolve it in some time . Thank you 👍

Aryan0699:
@GautamBytes 
For first comment i am fetching 500 always so that after dedup i have enough values to apply limit. If i fetch based on what is asked after dedup they might become less . We can reduce the limit i guess. We can have it adjusted based on the requested limit like ->[ limit: Math.min(500, parsed.data.limit * 5) ]. I think this would work ?
And i have capped the products to 20 so that max 41 filters will be formed

GautamBytes:
Yeah, dynamic overfetch makes sense. I’d just base it on the actual response cap so tiny requests don’t still over-query.

For the 20 product cap, that sounds fine if it’s applied before building the review filters.Also once pushed, please also add a small _meta hint when the cap is hit so callers know review lookup was partial.

Aryan0699:
@GautamBytes 
Done :+1:

GautamBytes (approved):
LGTM

Aryan0699:
A few test are remaining will add them in following PR

GautamBytes:
I think the best next step is either:
1. retitle/rescope this as UI + root MCP cleanup only, with invalid price behavior fixed, or
2. close/supersede this with a new shared-parser extraction PR based on the current `packages/shopstr-mcp/src/parse-tags.ts`. (after ur #543 is merged)

I prefer option 2 so the read-only MCP project remains the source for the agent-facing parser contract.

wdyt?

Aryan0699:
I think we can merge the #430  as the retitled/rescoped PR to UI + root MCP cleanup only, with invalid price behavior fixed and also this current PR . Later on I will create a shared new parser that would include the current MCP package as well ?
Because i think we can build on top of #430 later

GautamBytes:
> I think we can merge the #430 as the retitled/rescoped PR to UI + root MCP cleanup only, with invalid price behavior fixed and also this current PR . Later on I will create a shared new parser that would include the current MCP package as well ? Because i think we can build on top of #430 later

my bad this comment was for #430 , but yeah u can proceed as discussed

Aryan0699:
Hmm thought the same . Could you mention this there so that it would be easier for @calvadev @Abhishek-Punhani to follow up. Also what are you thoughts on handling the build issue

GautamBytes:
> Also what are you thoughts on handling the build issue

I think the shared parser should first be made package-neutral, with no root app `@/...` imports. Then we can either copy it into `packages/shopstr-mcp` in a prebuild step or move it into a tiny shared package.

PR #552, opened 2026-06-29
fix: cryptographically validate zap receipts instead of just polling relays

shivv23:
### Description

This PR hardens Shopstr's Zapsnag/NIP-57 payment validation so purchase and inventory flows no longer trust any relay-returned kind-9735 event by shape alone.

Before this change, a bogus relay event could be counted as a payment because receipt signatures, receipt signer identity, embedded zap requests, BOLT11 invoice amount, and preimage/payment-hash proof were not validated end to end.

### Changes

- Reworked `utils/nostr/zap-validator.ts` around an options-object API for `validateSingleReceipt()` and `validateZapReceipt()`.
- Verify receipt events with `verifyEvent()`, require kind `9735`, and require `receipt.pubkey === expectedReceiptSignerPubkey`.
- Treat the LNURL provider `nostrPubkey` as the expected receipt signer and zap recipient, per the LNURL/NIP-57 flow.
- Validate receipt and embedded zap-request `p` tags against the LNURL provider pubkey.
- Validate product `e` tags against the expected product id.
- Require and decode the receipt `bolt11` invoice, then use the invoice millisats as the canonical payment amount.
- Keep relay `amount` tags optional, but strict when present.
- Reject malformed zap-request amounts and invoice/request/receipt amount mismatches.
- Verify the returned WebLN preimage against the BOLT11 invoice payment hash when `expectedPreimage` is supplied.
- Reject conflicting optional LNURL tags when `expectedLnurl` is supplied.
- Keep the freshness window symmetric around payment start time while allowing inventory validation to skip freshness.
- Updated `components/ZapsnagButton.tsx` to resolve the seller profile Lightning Address, require NIP-57 support via `LightningAddress.nostrPubkey`, and pass the LNURL provider pubkey into purchase and inventory validation.
- Kept `product.pubkey` for seller profile lookup and encrypted shipping/order message recipient, so seller communication still goes to the listing author.
- Expanded unit coverage for validator edge cases and Zapsnag component wiring, including wrong receipt signer, missing/invalid `bolt11`, invoice amount mismatch, malformed zap-request amount, optional receipt amount mismatch, preimage/payment-hash validation, optional LNURL mismatch, and inventory validation against the LNURL provider key.

### Verification
- Temporary fake-money browser smoke in Chrome:
  - Local fake LNURL/WebLN/invoice/receipt/relay harness.
  - Real Shopstr `ZapsnagButton` UI and real receipt validator executed in browser.
  - Verified alert: `Order Placed & Verified! Preimage: 11111111...`
  - Verified invoice amount: `100000` msats = `100` sats.
  - Verified receipt validator result: `valid: true`, `errors: []`.
  - No mainnet or testnet sats were spent.

### DIFF
--- a/components/ZapsnagButton.tsx
+++ b/components/ZapsnagButton.tsx
@@ -25,7 +25,53 @@ import {
 import { generateSecretKey, getPublicKey } from "nostr-tools";
 import { SHOPSTRBUTTONCLASSNAMES } from "@/utils/STATIC-VARIABLES";
 import { ProductData } from "@/utils/parsers/product-parser-functions";
-import { validateZapReceipt, validateSingleReceipt } from "@/utils/nostr/zap-validator";
+import {
+  validateZapReceipt,
+  validateSingleReceipt,
+} from "@/utils/nostr/zap-validator";
+import type { NostrManager } from "@/utils/nostr/nostr-manager";
+
+type SellerZapContext = {
+  lightningAddress: LightningAddress;
+  zapRecipientPubkey: string;
+};
+
+async function resolveSellerZapContext(
+  nostrManager: NostrManager,
+  sellerPubkey: string
+): Promise<SellerZapContext> {
+  const profileFilter = { kinds: [0], authors: [sellerPubkey] };
+  const events = await nostrManager.fetch([profileFilter]);
+  let lud16 = "";
+
+  if (events.length > 0) {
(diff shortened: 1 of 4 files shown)


GautamBytes:
did some changes to make it more secure and complete . Also did end to end browser testing which is now passing. pinging @calvadev for final review and merge now!

calvadev (approved):

Issue #578, opened 2026-07-21
[BUG] Marketplace stalls for minutes on load when a relay is slow/unreachable - NostrManager connection handling bugs


Aryan0699:
### Description

NostrManager (our wrapper around nostr-tools' SimplePool) had three separate bugs that
together caused the marketplace to hang for 2-3 minutes on page load whenever any relay
in the list was slow or unreachable (eg. relay.nostr.band). 

The three bugs found:

1. `addRelay()` cached the promise from the *first* `pool.ensureRelay()` call in a
   variable and reused it forever. Since a rejected Promise stays rejected,
   once a relay failed to connect once, every future `relay.connect()` call
   instantly re-threw that same stale failure instead of retrying - so a relay's
   `sleeping` flag could never recover for the rest of the session.

2. `NostrManagerParams.connectionTimeout` defaulted to `undefined`. This flowed into
   nostr-tools' `AbstractRelay.connect({ timeout: undefined })`, which only sets an
   internal timeout if given - so with `undefined`, no timeout was ever
   set, and a dead relay's connection attempt relied on the browsers default TCP
   timeout (~2-3 minutes) instead of failing fast.

3. `subscribe()` did `await this.keepAlive(relays)` before calling
   `this.pool.subscribeMap(...)`. `keepAlive()` uses `Promise.all()` internally,
   which only resolves once *every* relay's connect attempt has settled - including
   failing ones. So even though nostr-tools' `subscribeMap()` already connects and
   subscribes to each relay independently and in parallel, our own wrapper never gave
   it the chance - it waited for the slowest relay to finish connecting before asking
   *any* relay for data.

### Current Behavior

On page load, if any relay in the configured relay list is slow or unreachable
(eg. relay.nostr.band):

- The first REQ to any relay does not go out until ~2-3 minutes after page load
  (visible in video from the networks tab in the inspect panel).
- The marketplace shows a loading spinner the entire time, even though other relays
  in the list connect within milliseconds.
- Once a relay fails to connect one time, it never successfully reconnects again for
  the rest of the session - every subsequent load/fetch pays the same delay against
  that relay.
- Eventually the app-level fetch timeout (utils/timeout.ts, default 60s) fires and
  throws, discarding any events that had already been collected from healthy relays. But in reality since the  events came after the timeout none of them were captured .

### Expected Behavior

- Fast relays should connect and return listings within their normal response time,
  regardless of whether any other relay in the list is slow or dead.
[...]

PR #579, opened 2026-07-21
Fix Nostr relay load resilience

GautamBytes:
Summary:
- Bound relay connection timeout to 4s
- Load from healthy relays without waiting on slow/dead relays
- Keep partial fetch results on timeout

Fixes - #578


Aryan0699:
Thanks  @GautamBytes :smile:

Aryan0699:
Hi @GautamBytes subscribe() is still waiting for keepAlive() to finish before returning the subscription. We already fixed the same thing in utils/nostr/nostr-manager.ts by running it in the background.I think we should apply the same change over here as well ?

PR #606, opened 2026-08-09
feat(hodl): implement hodl invoice escrow end-to-end (Phase 3)

kaihere14:
Implements Phase 3 (hodl invoice escrow) end to end, following the design confirmed with @GautamBytes  on 7/28 and 7/31 (arbiter/backend holds the preimage, provider-based abstraction since no Lightning node is chosen yet, short server-side timeout for buyer silence).

**HodlInvoiceProvider interface + mock** (`utils/lightning/`)
- Four-method interface: `createHoldInvoice`, `lookupInvoice`, `settleInvoice`, `cancelInvoice`, no identity checks inside any of them, that's a separate layer.
- `paymentHashFromPreimage()` hashes decoded bytes, not hex text.
- In-memory mock only. No real Lightning node (LND/LNbits/CLN) wired in yet, per Gautam's 7/31 direction to keep this provider-based until a node/service is picked.

**Nostr event schema** (`utils/nostr/hodl-escrow-records.ts`)
- Kind 30408 (buyer-confirm), 30409 (arbiter-release-decision), 30410 (dispute-raised, shared by either party, no separate kinds needed).
- Confirm/dispute/release events expose only `authorPubkey`, never a trusted role label. Role is always resolved by comparing against the order's stored pubkeys, never from a self-reported tag (tested against forged role/`p` tags).
- `fetchHodlConfirmEvents`, `fetchHodlReleaseEvents`, `fetchHodlDisputeEvents` fetch all candidates and dedupe per author, not per newest, so a late-timestamped forgery can't hide a genuine event.

**Order registration / commitment** (`hodl_escrow_orders` table, `pages/api/db/register-hodl-order.ts`)
- Server generates the preimage, derives the payment hash, creates the invoice, and only on success writes one atomic row with buyer/seller/arbiter pubkeys and the preimage.
- Preimage never appears in any API response, and is excluded even from internal comparison reads.
- Arbiter pubkey read from server config only, buyer/seller/arbiter distinctness enforced at write time, first-write-wins on `payment_hash` prevents pre-registration/squatting.

**Authorization layer** (`authorizeHodlConfirmEventForOrder`, `authorizeHodlReleaseEventForOrder`)
- Both throw rather than return false/null on failure, so a caller cannot reach settlement without a successful check.
- Success returns a branded type (`AuthorizedHodlConfirmation` / `AuthorizedHodlRelease`), enforced at compile time as well as runtime.
- Release authorization is anchored on the order row's stored `arbiter_nostr_pubkey`, not the live env var, so rotating the configured arbiter cannot retroactively transfer authority over already-open escrows.
- `order_mismatch` added as a distinct failure case, checked before the DB read.

**Settle endpoint** (`pages/api/lightning/settle-hodl-invoice.ts`)
- Client sends only `paymentHash`. Server independently fetches and authorizes confirm events from relays, never trusts anything the client forwards.
- Centralized redaction on all logs and error paths so the preimage can never leak, caught a real bug in its own error handling during testing.

**Dispute resolution** (`pages/api/lightning/resolve-hodl-dispute.ts`)
- Same shape as the settle endpoint. Fetches release-decision events independently, authorizes against the arbiter, branches to `settleInvoice` (release to seller) or `cancelInvoice` (release to buyer, no preimage read on this path at all).

**Status sync** (`utils/lightning/hodl-status-sync.ts`)
- Provider-based sync of `open to accepted to cancelled` via `lookupInvoice`, no relay involvement.
- Guarded with `SELECT ... FOR UPDATE` so concurrent syncs cannot race, and a terminal status (`settled`/`cancelled`) can never be downgraded by a stale read.

**Seller dispute timeout** (`utils/nostr/hodl-dispute-actionability.ts`)
- Adds `accepted_at`, stamped only on the `open to accepted` transition.
- Buyer disputes are always actionable. Seller disputes are gated on a timeout since `accepted_at`, returning `remainingSeconds` rather than an error when not yet actionable, since this is a legitimate wait state, not a failure.

### Known gaps and open items

- No real Lightning node is wired in. This is intentional per Gautam's 7/31 direction, the provider interface is built so LND, LNbits, CLN, or a hosted node can be plugged in later with no other code changes.
- Kinds 30408/30409/30410 are not yet in `CACHEABLE_EVENT_KINDS`. Current implementation reads directly from relays, so this is not a blocker. Adding server-side caching later requires updating that allowlist and the `config_events` CHECK constraint together.
- The seller dispute timeout is currently set to 4 hours. This number is under discussion with Gautam, specifically whether a short hold-invoice timeout is compatible with multi-day physical shipping, or whether confirmation timing should be decoupled from the hold invoice window entirely.
- No arbiter portal UI is included. This PR covers the backend flow (events, authorization, settlement, dispute resolution, timeout evaluation) only.

### Resolved or fixed issue

none (Phase 3 feature work, not tied to a specific filed issue)

### Screenshots

Not applicable, backend only.

### Affirmation

- [x] My code follows the CONTRIBUTING.md guidelines


GautamBytes:
Any screenshots to show its working as of now?

kaihere14:
> Any screenshots to show its working as of now?

No UI yet since this is backend only and no real lightning provider is wired in, its built provider based so LND/LNbits/CLN can plug in later without touchin the rest of the code For proof it works end to end i can run the full flow against the mock provider (register order, confirm, authorize, settle, status updates) and share the terminal output/test run if that helps show its actually workin

GautamBytes (changes requested):
Could you please do the following:

* Connect a real LND node instead of using only the mock provider.

* Connect HODL escrow to the Shopstr checkout, orders page, buyer/seller controls, and arbiter page.

* Calculate the order amount on the server. Keep the logic to not trust an amount sent by the buyer.

* After Shopstr settles the HODL invoice, pay the seller’s saved Lightning address. This payout needs safe retries so it cannot be lost or paid twice.

* Automatically watch LND for invoice status changes. It should also recover correctly after a server restart.

* Fix the accepted_at database column and use TIMESTAMPTZ.

* Before moving money, check that a real dispute exists, belongs to the correct order, and has passed the required waiting time.

* Return a proper server error when Nostr relays are unavailable instead of treating it like missing authorization.

* Encrypt the invoice preimage and keep dispute details private using NIP-59.

* Test the complete flow using Shopstr in a browser with Polar/regtest and a real LND connection. Please share 
screenshots showing payment, accepted status, buyer confirmation, seller payout, buyer refund, and arbiter resolution.

kaihere14:
> Could you please do the following:
> 
>     * Connect a real LND node instead of using only the mock provider.
> 
>     * Connect HODL escrow to the Shopstr checkout, orders page, buyer/seller controls, and arbiter page.
> 
>     * Calculate the order amount on the server. Keep the logic to not trust an amount sent by the buyer.
> 
>     * After Shopstr settles the HODL invoice, pay the seller’s saved Lightning address. This payout needs safe retries so it cannot be lost or paid twice.
> 
>     * Automatically watch LND for invoice status changes. It should also recover correctly after a server restart.
> 
>     * Fix the accepted_at database column and use TIMESTAMPTZ.
> 
>     * Before moving money, check that a real dispute exists, belongs to the correct order, and has passed the required waiting time.
> 
>     * Return a proper server error when Nostr relays are unavailable instead of treating it like missing authorization.
> 
>     * Encrypt the invoice preimage and keep dispute details private using NIP-59.
> 
>     * Test the complete flow using Shopstr in a browser with Polar/regtest and a real LND connection. Please share
>       screenshots showing payment, accepted status, buyer confirmation, seller payout, buyer refund, and arbiter resolution.

Sure, I will completely set up and will ping you whenever it's completed. It would require me some time to complete all these steps

GautamBytes:
@calvadev , i have tested the pr and it seems to work fine now.  u can check here - https://docs.google.com/document/d/1piLnupYkzvb3Hyq_rt1LZ8uDRvm4IWY6ozOISujNMD4/edit?usp=sharing

GautamBytes (approved):
lgtm now

PR #616, opened 2026-08-16
Add NIP-50 support to search-products tool - MCP server

Aryan0699:
### Description
This PR adds NIP50 support for search_products tool of MCP server

- Added NIP-50 (search) relay querying - search_products now queries NIP-50-capable relays in parallel with the existing keyword/tag scan and merges matches into the response.
- Separated NIP-50 from normal relay pagination - normal and NIP-50 events are tracked independently so sparseBoundary and cursor calculations only use the normal stream.
- Reserved response slots for NIP-50 matches (fixed at 5) guarantees NIP-50 results get space in the response alongside normal matches.
- Tagged NIP-50 results - products returned through NIP-50 are marked with matchedVia: `nip50`.
- NIP-50 queries run for cursor-based requests instead of only the first page.
- Each page increases the NIP-50 relay limit and excludes identities already present in the pagination cursor.
- Prevented duplicate results - NIP-50 candidates already returned by the normal stream are excluded.
- Applied the the sparse-window cursor fix so that things are working perfectly
- NIP-50 results are not limited to a fixed NIP50_RESERVED_SLOTS allocation; unused response capacity from normal results is  reallocated to additional NIP-50 matches.
 - NIP-50 content matching added through event.content
 
### Resolved or fixed issue
`none`  

### Video  
https://drive.google.com/file/d/1nHKS0eolVLptf4679oz5p2sr11C7Z1Cm/view?usp=sharing

### Affirmation

### DIFF
--- a/packages/shopstr-mcp/.env.example
+++ b/packages/shopstr-mcp/.env.example
@@ -2,6 +2,9 @@
 # wss://nos.lol,wss://relay.damus.io,wss://purplepag.es
 SHOPSTR_MCP_RELAYS=wss://nos.lol,wss://relay.damus.io,wss://purplepag.es
 
+# Curated NIP-50 search relays used only for search_products keyword searches.
+SHOPSTR_MCP_NIP50_SEARCH_RELAYS=wss://relay.nostr.band,wss://nostr.wine,wss://relay.noswhere.com,wss://search.nos.today,wss://antiprimal.net,wss://relay.ditto.pub
+
 
 # Log level for json logs written to stderr.
 # levels : error, warn, info, debug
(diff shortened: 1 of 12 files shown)


GautamBytes (approved):

Aryan0699:
@GautamBytes Thanks for fixing the content field gap here itself.
Should i fix the other limitation here itself and also should i add the content field to the response of the get-product-details tool . If yes, then we would need to truncate the content size since there is no limit on it so beyond a particular word limit we can do  ... ?

GautamBytes:
> @GautamBytes Thanks for fixing the content field gap here itself. Should i fix the other limitation here itself and also should i add the content field to the response of the get-product-details tool . If yes, then we would need to truncate the content size since there is no limit on it so beyond a particular word limit we can do ... ?

yeah that would be beneficial , u can do in this pr only

